PR #28, opened 2026-04-04
BREAKING: fillTemplates() returns { result, unresolvedFields }

### LINKED ISSUE #27: Enhance fillTemplates(): allowUnresolved mode + return unfilled field names
## Context

Platforms using SCORE for annotation need two-pass template
expansion:

1. **First pass:** expand researcher templates (broadcast, field
   substitution), leaving platform-provided placeholders
   (`${clipUrl}`, etc.) unresolved. Extract treatment names for
   task randomization.
2. **Second pass:** for each treatment × clip set, fill the
   remaining placeholders with clip-specific values using
   `additionalFields`.

Both passes should use the same `fillTemplates()` function with
its iterative nested-field resolution. The difference is what
happens when placeholders remain at the end.

## Proposal

Add `allowUnresolved` option. When true, instead of throwing on
remaining `${...}` placeholders, return them alongside the
result:

```typescript
// Current signature (unchanged for backward compat)
export function fillTemplates(options: {
  obj: any;
  templates: any[];
  additionalFields?: Record<string, any>;
}): any;

// New overload with allowUnresolved
export function fillTemplates(options: {
  obj: any;
  templates: any[];
  additionalFields?: Record<string, any>;
  allowUnresolved: true;
}): { result: any; unresolvedFields: string[] };
```

### Behavior with `allowUnresolved: true`

1. Expand templates recursively (same as today)
2. Apply `additionalFields` if provided (same as today)
3. Iterate until no more fields are being resolved (same
   convergence logic as today's template-remaining loop)
4. Instead of throwing on remaining `${...}` placeholders,
   collect their names and return them alongside the result

This means a platform can use the same function for both passes:

```typescript
// Pass 1: expand researcher templates, get treatment list
const { result: expanded, unresolvedFields } = fillTemplates({
  obj: rawTreatmentFile,
  templates: rawTreatmentFile.templates ?? [],
  allowUnresolved: true,
});
// unresolvedFields = ["clipUrl", "clipStartAt", "clipStopAt"]
// expanded.treatments[] has names, with ${clipUrl} etc. intact

// Pass 2: for each treatment × clip set, fill clip fields
const { result: resolved, unresolvedFields: remaining } = fillTemplates({
  obj: expanded.treatments[i],
  templates: [],
  additionalFields: {
    clipUrl: "solo_recordings/.../pos0.mp4",
    clipStartAt: 12.5,
    clipStopAt: 45.0,
  },
  allowUnresolved: true,
});

if (remaining.length > 0) {
  throw new Error(`Unresolved fields: ${remaining.join(", ")}`);
}
```

Or for the simple one-pass case (no change needed):

```typescript
// Throws if any fields remain (backward compatible)
const result = fillTemplates({
  obj: rawTreatmentFile,
  templates: rawTreatmentFile.templates ?? [],
  additionalFields: clipFields,
});
```

## Implementation

In `fillTemplates.ts`:

1. Add `allowUnresolved?: boolean` to the options type
2. After the existing iterative expansion + `additionalFields`
   substitution:
   - Collect remaining `${...}` placeholder names (already done
     on line 246)
   - If `allowUnresolved` is true: return
     `{ result: newObj, unresolvedFields: [...missingFields] }`
   - If `allowUnresolved` is false/undefined: throw as today
     (backward compatible)

3. Export `substituteFields` so platforms can also do targeted
   field substitution outside of `fillTemplates` if needed.

## Why not just getUnresolvedFields() + additionalFields?

`getUnresolvedFields()` (#23) tells you what fields are missing
but doesn't give you the partially-expanded result. You'd have
to call `fillTemplates` twice — once to discover fields (which
throws), and once with all fields provided. With
`allowUnresolved`, you get the intermediate result in one call
and can inspect/use it before the second pass.

## Supersedes

This supersedes the simpler `allowUnresolved` proposal in the
original version of this issue. The key addition is returning
`unresolvedFields` instead of silently swallowing them, and
ensuring `additionalFields` works in the same call for the
second pass.

JamesPHoughton:
## Summary

`fillTemplates()` now always returns `{ result, unresolvedFields }` instead of the bare expanded object. New `allowUnresolved` option enables two-pass template expansion.

## Breaking change

```typescript
// Before
const expanded = fillTemplates({ obj, templates });

// After
const { result: expanded } = fillTemplates({ obj, templates });
```

## New: allowUnresolved

```typescript
// Pass 1: expand researcher templates, leave platform fields
const { result: expanded, unresolvedFields } = fillTemplates({
  obj: rawTreatmentFile,
  templates: rawTreatmentFile.templates ?? [],
  allowUnresolved: true,
});
// unresolvedFields = ["clipUrl", "clipStartAt", "clipStopAt"]

// Pass 2: fill platform fields
const { result: resolved } = fillTemplates({
  obj: expanded,
  templates: [],
  additionalFields: { clipUrl: "clip1.mp4", clipStartAt: 12.5, clipStopAt: 45.0 },
});
```

## Changes

- `fillTemplates()` signature: returns `{ result: any; unresolvedFields: string[] }`
- `allowUnresolved` option: `true` returns partial result, `false` (default) throws
- `getUnresolvedFields()` deprecated in favor of `fillTemplates({ allowUnresolved: true })`
- All 24 existing test call sites updated to destructure `{ result }`
- 7 new tests for allowUnresolved + two-pass pattern

## Test plan

- [x] All existing tests pass with new return type (166 vitest)
- [x] allowUnresolved returns partial result with field names
- [x] Two-pass expansion: broadcast + additionalFields
- [x] Default behavior still throws on unresolved
- [x] Playwright CT tests unaffected (146 pass)

Closes #27.

### DIFF
--- a/src/schemas/treatment.ts
+++ b/src/schemas/treatment.ts
@@ -18,8 +18,11 @@ function isValidRegex(pattern: string): boolean {
 
 // ------------------ Names, descriptions, and files ------------------ //
 
-const fieldPlaceholderSchema = z.string().regex(/\$\{[a-zA-Z0-9-_ ]+\}/, {
-  message: "Field placeholder must be in the format `${fieldKey}`",
+// Field placeholder pattern: only alphanumeric + underscore.
+// Must match FIELD_PLACEHOLDER_REGEX in templates/fillTemplates.ts
+const fieldPlaceholderSchema = z.string().regex(/\$\{[a-zA-Z0-9_]+\}/, {
+  message:
+    "Field placeholder must be in the format `${fieldKey}` (alphanumeric and underscores only)",
 });
 
 // Names should have properties:
@@ -349,15 +352,11 @@ export const discussionSchema = z
 export type DiscussionType = z.infer<typeof discussionSchema>;
 
 // ------------------ Template contexts ------------------ //
-const templateFieldKeysSchema = z // todo: check that the researcher doesn't try to overwrite the dimension keys (d0, d1, etc.)
+const templateFieldKeysSchema = z
   .string()
-  // .regex(/^(?!d[0-9]+)[a-zA-Z0-9_]+$/, {
-  //   message:
-  //     "String must only contain alphanumeric characters and underscores, and not overwrite the broadcast dimension keys `d0`, `d1`, etc.",
-  // })
-  .regex(/^(?!d[0-9]+$)([a-zA-Z0-9-_ ]+|\$\{[a-zA-Z0-9_]+\})$/, {
+  .regex(/^(?!d[0-9]+$)([a-zA-Z0-9_]+|\$\{[a-zA-Z0-9_]+\})$/, {
     message:
-      "Field key must be alphanumeric, may include underscores, dashes, or spaces, or be in the format `${fieldKey}` without conflicting with reserved keys (e.g., `d0`, `d1`, etc.).",
+      "Field key must be alphanumeric with underscores only (no hyphens or spaces), or a placeholder `${fieldKey}`. Cannot conflict with reserved keys (d0, d1, etc.).",
   })
   .min(1)
   .superRefine((val, ctx) => {
--- a/src/templates/fillTemplates.test.ts
+++ b/src/templates/fillTemplates.test.ts
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-return */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
 
 import { expect, test } from "vitest";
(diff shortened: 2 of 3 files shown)


JamesPHoughton:
Addressed both Copilot review comments:

**1. Regex alignment:** Tightened field key pattern to `[a-zA-Z0-9_]` only — dropped hyphens and spaces. Extracted `FIELD_PLACEHOLDER_REGEX` as a shared constant in `fillTemplates.ts`. The schema (`fieldPlaceholderSchema` and `templateFieldKeysSchema` in `treatment.ts`) now uses the same restricted character set. Both locations have comments pointing to each other to prevent drift.

**2. `getUnresolvedFields` is now a thin wrapper:** Reimplemented as a one-liner that calls `fillTemplates({ ..., allowUnresolved: true })` and returns `unresolvedFields`. No duplicated logic.

PR #147, opened 2026-04-17
Address hygiene fixes from issue #99

### LINKED ISSUE #99: Non-breaking hygiene fixes
Collecting small, non-breaking cleanup items surfaced from reviews. Each is independent and can be addressed in any order.

## Interface / type hygiene

- [ ] **`renderSharedNotepad` type mismatch** — `StagebookProvider.tsx:43` declares `(config: { padName: string })` but `Prompt.tsx:162` actually passes `{ padName, defaultText, rows }`. Platforms silently lose the default text and row count. Update the interface type to include `defaultText?: string` and `rows?: number`.

- [ ] **`useResolvedParams` naming** — `Element.tsx:19` defines a plain function (no hook calls) with a `use` prefix, which is misleading. If it ever added internal hook calls, the conditional invocation at line 329 (`case "qualtrics"`) would become a Rules of Hooks violation. Rename to `resolveParams` or similar.

## Platform contract cleanup

- [ ] **Normalize `get()` scope values in the provider** — The provider currently passes `"any"` and `"percentAgreement"` through to the platform's `get()`. The semantic distinction only matters in `evaluateCondition()` *after* the values come back. The provider could map both to `"all"` before calling `get()`, reducing the platform's scope vocabulary to `"player" | "shared" | "all" | numeric string`.

- [ ] **Document `progressLabel` uniqueness requirement** — `progressLabel` is used both as metadata in saved records (`Element.tsx:108`) and as a component of auto-generated storage keys (`Element.tsx:184` fallback prompt name). Collisions cause silent data overwrites. Add documentation in `platform-requirements.md`.

- [ ] **`participantInfo` storage asymmetry** — `participantInfo.name` resolves to bare key `"name"` while `connectionInfo.country` resolves to key `"connectionInfo"` with path `["country"]` (`reference.ts:36-39`). Surprising for platform implementers. Consider normalizing to match `connectionInfo`/`browserInfo` pattern. Would be a breaking change to the DSL — needs discussion.

## Defense in depth

- [ ] **`evaluateConditions.ts:21` prototype key collision** — The `counts` record uses `{}` which inherits from `Object.prototype`. A response value of `"constructor"` produces `NaN` counts. Use `Object.create(null)` or `Map<string, number>`.

- [ ] **`mapPosition` in `context.ts:98`** — `Number(positionArg)` accepts `Infinity`, `-Infinity`, and negative numbers. Add `Number.isFinite(num) && num >= 0 && Number.isInteger(num)` guard.

- [ ] **`getNestedValueByPath` prototype property denylist** — `reference.ts:1-11` traverses objects by arbitrary path segments with no filtering. Adding a denylist for `__proto__`, `constructor`, `prototype` is a low-cost defense-in-depth measure. (Read-only in this context, so exploitability is limited, but the `StateInspector` combines path traversal with writes.)

JamesPHoughton:
Closes #99.

All 8 items from the issue are addressed. Breaking changes and design choices follow the decisions posted on the issue by @jamesphoughton.

## Interface / type hygiene

- `renderSharedNotepad` in `StagebookContext` now declares `defaultText?: string` and `rows?: number` so its type matches what `Prompt.tsx` actually passes.
- `useResolvedParams` → `resolveParams`. It's a plain function (no hook calls), so the `use` prefix was misleading and would have turned the conditional call site in the `qualtrics` case into a Rules-of-Hooks violation the moment it ever grew internal hooks.

## Platform contract cleanup

- `StagebookProvider.resolve()` now maps `"any"` and `"percentAgreement"` to `"all"` before calling `platform.get()`. Platforms now see a minimal scope vocabulary: `"player" | "shared" | "all" | numeric string`. Condition-evaluation semantics stay in `evaluateCondition()`, where the original position is still available on the condition object. The viewer's `mapPosition` is simplified accordingly.
- `participantInfo.<field>` now resolves to storage key `"participantInfo"` with nested path `[field, ...]`, matching `connectionInfo`/`browserInfo`. **Breaking change** — platforms must store participant info as a nested object under `participantInfo` instead of flat keys. Schema validation and `platform-requirements.md` are updated to match.
- Added a `progressLabel` uniqueness section to `docs/engineer/platform-requirements.md` explaining the hard invariant and why collisions silently overwrite data.

## Defense in depth

- `evaluateConditions.ts`: the `percentAgreement` counts map is now `Object.create(null)`, so response values like `"constructor"` or `"toString"` no longer inherit from `Object.prototype` and produce `NaN` counts. Regression test added.
- `context.ts` `mapPosition`: the numeric branch now requires `Number.isFinite(num) && Number.isInteger(num) && num >= 0`. `"Infinity"`, `"-Infinity"`, `"-1"`, and `"1.5"` fall through to the current position. Regression test added.
- `getNestedValueByPath`: path segments `__proto__`, `constructor`, `prototype` are denied. Read-only in this context, but `StateInspector` combines path traversal with writes, so it's a cheap defence-in-depth. Regression test added.

## Test plan

- [x] `npm test` — 685 tests across three workspaces, all pass
- [x] `npm run lint` — zero warnings
- [x] `npm run build -w stagebook` — succeeds, dts emitted

### DIFF
--- a/apps/viewer/src/lib/context.test.ts
+++ b/apps/viewer/src/lib/context.test.ts
@@ -62,12 +62,15 @@ describe("createViewerContext", () => {
       expect(values).toEqual([{ value: "a" }, { value: "b" }]);
     });
 
-    it("get with 'any' returns raw values from all positions", () => {
-      const { store, ctx } = makeContext();
-      store.save("prompt_q1", { value: "a" }, "player", 0, 0);
-      store.save("prompt_q1", { value: "b" }, "player", 1, 0);
-      const values = ctx.get("prompt_q1", "any");
-      expect(values).toEqual([{ value: "a" }, { value: "b" }]);
+    it("get with non-finite numeric scope falls back to current position", () => {
+      // Number("Infinity"), Number("-Infinity"), and negative numbers
+      // must not be accepted as position indices.
+      const { store, ctx } = makeContext({ position: 0 });
+      store.save("prompt_q1", { value: "own" }, "player", 0, 0);
+      expect(ctx.get("prompt_q1", "Infinity")).toEqual([{ value: "own" }]);
+      expect(ctx.get("prompt_q1", "-Infinity")).toEqual([{ value: "own" }]);
+      expect(ctx.get("prompt_q1", "-1")).toEqual([{ value: "own" }]);
+      expect(ctx.get("prompt_q1", "1.5")).toEqual([{ value: "own" }]);
     });
   });
 
--- a/apps/viewer/src/lib/context.ts
+++ b/apps/viewer/src/lib/context.ts
@@ -46,7 +46,7 @@ export function createViewerContext(
       if (typeof mapped === "number" || mapped === "shared") {
         return store.lookup(key, mapped);
       }
-      // "all", "any", "percentAgreement" → return from all player positions
+      // "all" → return from all player positions
       return store.lookup(key);
     },
 
@@ -88,15 +88,11 @@ function mapPosition(
   if (positionArg === "shared") {
     return "shared";
   }
-  if (
-    positionArg === "all" ||
-    positionArg === "any" ||
-    positionArg === "percentAgreement"
-  ) {
+  if (positionArg === "all") {
     return "all";
   }
   const num = Number(positionArg);
-  if (!Number.isNaN(num)) {
+  if (Number.isFinite(num) && Number.isInteger(num) && num >= 0) {
     return num;
   }
   return currentPosition;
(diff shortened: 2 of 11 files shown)


JamesPHoughton:

All three Copilot review points are addressed in fd4b9b0:

- `resolveParams` comment in `Element.tsx` reworded to be element-agnostic (it's called from both `trackedLink` and `qualtrics`).
- `getReferenceKeyAndPath` now throws when a flat-namespace reference (`urlParams` / `connectionInfo` / `browserInfo` / `participantInfo`) has no path segment, matching `referenceSchema`. Regression test added.
- Fixed the malformed example in the `referenceSchema` error message (stray trailing `.` instead of closing `'`).

CI is green on all four jobs.

---

Issue #103, opened 2026-04-15
PlaybackProvider register/unregister causes infinite re-render loop


JamesPHoughton:
## Problem

\`PlaybackProvider\` has an infinite re-render loop when a \`MediaPlayer\` is present on the page. The loop is:

1. MediaPlayer mounts → calls \`register(name, handle)\` in a useEffect
2. \`register\` calls \`setHandles(prev => new Map(prev).set(name, handle))\`
3. \`handles\` changes → \`value\` useMemo recalculates (new object ref)
4. Context change → all consumers re-render → MediaPlayer's useEffect cleanup fires → calls \`unregister\`
5. \`unregister\` calls \`setHandles(prev => ...)\` → \`handles\` changes
6. Context change → MediaPlayer re-renders → useEffect fires → calls \`register\` → back to step 2

## Stack trace

\`\`\`
Maximum update depth exceeded.
dispatchSetState @ react-dom_client.js
(anonymous) @ stagebook_components.js:41435  ← setHandles in PlaybackProvider
(anonymous) @ stagebook_components.js:41454
commitHookEffectListUnmount @ react-dom_client.js
\`\`\`

## Root cause

\`register\` and \`unregister\` create a new \`Map\` on every call (\`new Map(prev)\`), which always changes the \`handles\` reference, which always changes the \`value\` memo, which always triggers consumer re-renders.

## Suggested fix

Check if the handle being registered is the same reference already in the map:

\`\`\`typescript
const register = useCallback((name: string, handle: PlaybackHandle) => {
  setHandles((prev) => {
    if (prev.get(name) === handle) return prev; // no change
    return new Map(prev).set(name, handle);
  });
}, []);
\`\`\`

Or use a ref for the handles map and only use state for triggering renders when actually needed.

## Context

Discovered in the annotator platform. The loop fires on any page that renders a \`mediaPlayer\` element. It triggers React's "Maximum update depth exceeded" error. Without a MediaPlayer, there's no loop.

PR #150, opened 2026-04-17
Apply ref pattern to unstable callback dependencies (#105)

### LINKED ISSUE #105: Audit unstable callback dependencies in useEffect hooks
## Background

While fixing the `PlaybackProvider` infinite re-render loop in #103, we identified the root cause: unstable references (callbacks, context objects) used as `useEffect` dependencies cause effects to re-fire every render. When those effects trigger state changes, you get an infinite loop.

The fix for #103 used two techniques:
1. **Bail-out checks** in state updaters — return `prev` when nothing actually changed
2. **Ref the context** — store the context in a `useRef` so the effect doesn't depend on the context object changing

A search of the codebase found several other locations with the same risky pattern: **unstable callbacks used as effect dependencies**.

### High risk — effect depends on potentially unstable callback

| File | Line | Dependency | Risk |
|------|------|------------|------|
| `Qualtrics.tsx` | 52 | `onComplete` | Effect calls `onComplete()` inside — if the reference is unstable, it loops |
| `MediaPlayer.tsx` | 484 | `onComplete` | `handleEnded` callback recreated when `onComplete` changes, cascading to event handlers |
| `MediaPlayer.tsx` | 572 | `onComplete` | `handleTimeUpdate` fires every frame — unstable dep means constant recreation |
| `AudioElement.tsx` | 36 | `save` | Effect sets up event listeners; unstable `save` causes listener churn |
| `Timeline.tsx` | 218 | `save` | Save effect re-runs every render if `save` is unstable |

### Moderate risk — context function as dependency

| File | Line | Dependency | Risk |
|------|------|------------|------|
| `StagebookProvider.tsx` | 172 | `getTextContent` | `useTextContent` refetches content on every render if `getTextContent` is recreated |
| `HelpPopover.tsx` | 60 | `onClose` | Event listeners re-registered every render |

## Whether these actually loop today

These are **latent risks**, not necessarily active bugs. Whether they loop depends on how the parent component provides the callback:
- If the parent wraps it in `useCallback` → stable, no problem
- If the parent passes an inline arrow function → unstable, effect re-fires every render
- If the effect also triggers a state change → infinite loop

The `Element.tsx` router and `StagebookProvider` are the main parents providing `save`, `onComplete`, and `getTextContent`. If those are stable today, the components work fine — but they're fragile, and a future refactor could break them.

## Suggested mitigations

1. **Ref pattern** (used in #103 fix): Store the callback in a `useRef` and read `ref.current` inside the effect. The effect no longer depends on the callback identity.

   ```typescript
   const saveRef = useRef(save);
   saveRef.current = save;

   useEffect(() => {
     // use saveRef.current instead of save
   }, [/* other deps, but not save */]);
   ```

2. **Memoize at the source**: Ensure `save`, `onComplete`, and `getTextContent` are wrapped in `useCallback` where they're defined (in `Element.tsx`, `StagebookProvider`, etc.).

3. **Bail-out checks in state updaters**: When a setState call creates a new collection (`new Map()`, `[...prev]`, `{...prev}`), check if the content actually changed before returning a new reference.

Approach 1 (ref pattern) is the most defensive — it works regardless of what the parent does. Approach 2 is cleaner but requires discipline across all call sites.

JamesPHoughton:
Closes #105.

## Summary

Audit and mitigation for the latent render-loop risks flagged after the #103 fix. Every site the issue called out now stores the unstable callback in a `useRef`, reads `ref.current` inside the effect/handler, and drops the callback from the dependency array. Behavior is unchanged when parents already stabilize callbacks — components are now also defensive against parents that don't.

## Sites fixed

**High risk — effect depends on potentially unstable callback**

- `Qualtrics.tsx` (l. 52) — effect previously re-added its window `"message"` listener on every render when `save` / `onComplete` were unstable.
- `MediaPlayer.tsx` (ll. 450, 484, 572) — `recordEvent`, `handleEnded`, `handleTimeUpdate`, and the YouTube `onPause` / `onEnded` handlers depended on `save` / `onComplete` / `getElapsedTime`. `handleTimeUpdate` fires every animation frame during playback — churning its closure cascaded to every JSX handler tied to the `<video>` element.
- `AudioElement.tsx` (l. 36) — the single effect set up a `canplaythrough` listener on a fresh `Audio` instance; unstable `save` tore it down every render.
- `Timeline.tsx` (l. 218) — selection-save effect re-ran on every render when `save` was unstable; could double-save across renders.

**Moderate risk — context function as dependency**

- `StagebookProvider.tsx` / `useTextContent` (l. 172) — refetched content every render if the platform rebuilt its context without `useCallback`-ing `getTextContent`.
- `HelpPopover.tsx` (l. 60) — re-registered document `keydown` / `mousedown` listeners every render on unstable `onClose`.

## Approach

Used the ref pattern recommended by the issue (approach 1), matching the #103 fix:

```ts
const saveRef = useRef(save);
saveRef.current = save;

useEffect(() => {
  // read saveRef.current inside
}, [/* other deps, but not save */]);
```

This is the most defensive option — it works regardless of whether the parent memoizes the callback.

## Tests (red → green)

New regression tests for each tractable site. Each drives multiple re-renders with fresh inline callback identities and asserts the effect sets up exactly once. For `Qualtrics` and `HelpPopover`, a companion test confirms the *latest* callback is still invoked when the event eventually arrives (so refs are read, not closed-over).

- `packages/stagebook/src/components/elements/Qualtrics.test.tsx` (new)
- `packages/stagebook/src/components/elements/AudioElement.test.tsx` (new)
- `packages/stagebook/src/components/elements/timeline/HelpPopover.test.tsx` (new)
- `packages/stagebook/src/components/StagebookProvider.test.tsx` — added `useTextContent` case with a rebuilt context on every render

Each new test was observed to **fail on the unpatched tree** (6 setup calls instead of 1, or 6 fetches instead of 1) and pass after the ref-pattern edits.

## Test plan

- [x] `npm test` — stagebook 491, viewer 59, vscode 136 all pass
- [x] `npm run lint` — clean (zero warnings)
- [x] `npm run build -w stagebook` — succeeds
- [ ] Playwright CT — runs in CI (not available in sandbox)

## Notes

- `MediaPlayer.handleTimeUpdate` and `handleEnded` keep their `recordEvent` / `submitOnComplete` / `stopAt` / `cues` deps (still needed) but drop `onComplete`.
- `MediaPlayer.recordEvent` keeps `name` / `url` / `startAt` / `stopAt` / `saveKey` and drops `save` + `getElapsedTime`.
- `Timeline` effect keeps `state.selections` / `isDragging` / `name` and drops `save`.
- `useTextContent` effect keeps `path` and drops `getTextContent`.
- `HelpPopover` effect now has `[]` deps (everything it reads is either stable state or a ref).

### DIFF
--- a/docs/engineer/platform-requirements.md
+++ b/docs/engineer/platform-requirements.md
@@ -99,6 +99,33 @@ Stagebook components are tested against modern browsers. The platform should ver
 
 Mobile devices are not supported for interactive experiments with video/audio. The platform should detect and block mobile browsers during onboarding.
 
+### Callback Stability (recommended)
+
+The methods the platform provides on `StagebookContext` — `get`, `save`, `getTextContent`, `getAssetURL`, `getElapsedTime`, `submit`, and the optional `render*` slots — are safest when their identities are stable across renders.
+
+Stagebook internally protects against unstable callback identities (it stores these references in refs so effects and event listeners aren't torn down and re-registered each render). But stable callbacks still help in a few ways:
+
+- Prompt components that read state via `get()` re-run resolution whenever the context identity changes. A stable context object (same `get` identity across renders) lets React skip work.
+- Platform-side React devtools and profiling are easier to read when the context value isn't churning.
+- If a future Stagebook component is written without the defensive ref pattern, stable callbacks are what keeps it correct.
+
+The simplest way to get stable references is to wrap each method in `useCallback` (or define it outside the render, or on a class) and memoize the final context object with `useMemo`:
+
+```tsx
+const save = useCallback((key, value, scope) => { /* ... */ }, [/* store deps */]);
+const get = useCallback((key, scope) => { /* ... */ }, [/* store deps */]);
+const getTextContent = useCallback((path) => fetch(resolve(path)), []);
+
+const ctx = useMemo<StagebookContext>(() => ({
+  get, save, getTextContent, getAssetURL, getElapsedTime, submit,
+  progressLabel, playerId, position, playerCount, isSubmitted,
+}), [get, save, getTextContent, /* ... */]);
+
+return <StagebookProvider value={ctx}>{children}</StagebookProvider>;
+```
+
+This is a recommendation, not a hard requirement — Stagebook will behave correctly either way.
+
 ---
 
 ## 2. Stage Orchestration (required)
--- a/packages/stagebook/src/components/StagebookProvider.test.tsx
+++ b/packages/stagebook/src/components/StagebookProvider.test.tsx
@@ -326,4 +326,60 @@ describe("useTextContent", () => {
     expect(result.current.error).toBeUndefined();
     unmount();
   });
+
+  // Regression for #105: useTextContent previously listed `getTextContent`
+  // in its effect deps. If the platform rebuilds its StagebookContext each
+  // render (very common) and does not wrap `getTextContent` in `useCallback`,
+  // the effect re-fires every render — refetching content on every parent
+  // state change. The ref pattern fixes this.
+  test("does not refetch when getTextContent identity changes but path is stable", async () => {
+    const container = document.createElement("div");
+    let root: Root;
+    const result = { current: undefined as unknown as TextContentResult };
+    const getTextContent = vi.fn(() => Promise.resolve("# Hello"));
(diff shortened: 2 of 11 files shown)


JamesPHoughton:

---

PR #151, opened 2026-04-18
Accept `timeline.*` references in referenceSchema (#131)

### LINKED ISSUE #131: Schema rejects `timeline.*` references in conditions
## Summary

The VS Code extension (and any consumer of `referenceSchema`) flags valid `timeline.<name>` references as invalid, producing:

> Invalid reference type "timeline", need to be in form of a valid reference type such as 'survey', 'submitButton', 'qualtrics', 'discussion', 'participantInfo', 'prompt', 'urlParams', 'connectionInfo', or 'browserInfo' followed by a . and name or path.

Repro:

```yaml
- type: timeline
  source: story
  name: storySegment
  selectionType: range
  multiSelect: false

- type: submitButton
  conditions:
    - reference: timeline.storySegment
      comparator: exists
```

## Root cause

`referenceSchema` in [packages/stagebook/src/schemas/treatment.ts](packages/stagebook/src/schemas/treatment.ts#L445-L504) has a `switch (givenType)` that predates the timeline element. It never got a `case "timeline"` branch, so timeline references fall into the `default` branch and are rejected at validation time.

The runtime resolver in [packages/stagebook/src/utils/reference.ts:26](packages/stagebook/src/utils/reference.ts#L26) **does** include `"timeline"` — so the reference would resolve fine if it got past schema validation. The schema and the resolver are out of sync.

The default-case error message is also stale: it omits both `timeline` and `trackedLink` (which is already supported by both the schema switch and the resolver).

## Patch

1. Add `case "timeline":` to the `discussion`/`participantInfo`/`prompt`/`trackedLink` branch in [treatment.ts:472-484](packages/stagebook/src/schemas/treatment.ts#L472-L484) — same shape (name required, path optional).
2. Update the `default`-case error message at [treatment.ts:500](packages/stagebook/src/schemas/treatment.ts#L500) to include `timeline` and `trackedLink`.
3. Add test cases in [packages/stagebook/src/schemas/treatment.test.ts](packages/stagebook/src/schemas/treatment.test.ts):
   - `timeline.storySegment` (whole array)
   - `timeline.storySegment.0.start` (nested path)
   - negative: `timeline` alone should still fail with "name must be provided"

## Notes

- Docs already describe the intended reference syntax at [docs/researcher/elements.md:353](docs/researcher/elements.md#L353) and the resolver has test coverage at [packages/stagebook/src/utils/reference.test.ts:61](packages/stagebook/src/utils/reference.test.ts#L61) — only the schema is missing.
- Worth auditing whether other newer reference types have the same schema/resolver drift.

JamesPHoughton:
Closes #131.

## Summary

The `referenceSchema` switch in `packages/stagebook/src/schemas/treatment.ts` predated the `timeline` element, so valid `timeline.*` references fell through to the `default` branch and were rejected at validation time — even though the runtime resolver (`packages/stagebook/src/utils/reference.ts`) already supports them.

## Changes

- Add `case "timeline":` alongside `discussion`/`prompt`/`trackedLink` in `referenceSchema` (name required, optional path — matches how timeline is resolved).
- Refresh the `default`-case error message to include `timeline` and `trackedLink` (`trackedLink` was already a valid case but missing from the list).
- Three new unit tests in `treatment.test.ts`:
  - `timeline.storySegment` — bare reference, passes
  - `timeline.storySegment.0.start` — nested path, passes
  - `timeline` alone — still fails with "name must be provided"

## Audit of other reference types

Compared the schema's `switch` cases against the resolver's `if/else if` chain. The schema handles every type the resolver handles *except* `timeline`, which this PR fixes. No other drift.

## Test plan

- [x] `npm test` — stagebook 524 (was 521, +3), viewer 68, vscode 136, all pass
- [x] `npm run lint` — clean

### DIFF
--- a/packages/stagebook/src/schemas/treatment.test.ts
+++ b/packages/stagebook/src/schemas/treatment.test.ts
@@ -59,6 +59,28 @@ test("reference tracked link with name", () => {
   expect(result.success).toBe(true);
 });
 
+test("reference timeline with name", () => {
+  const reference = "timeline.storySegment";
+  const result = referenceSchema.safeParse(reference);
+  if (!result.success) console.log(result.error);
+  expect(result.success).toBe(true);
+});
+
+test("reference timeline with nested path", () => {
+  const reference = "timeline.storySegment.0.start";
+  const result = referenceSchema.safeParse(reference);
+  if (!result.success) console.log(result.error);
+  expect(result.success).toBe(true);
+});
+
+test("reference timeline with no name", () => {
+  const reference = "timeline";
+  const result = referenceSchema.safeParse(reference);
+  if (!result.success)
+    console.log(result.error.message, "\npath:", result.error.path);
+  expect(result.success).toBe(false);
+});
+
 // ----------- Condition Schema ------------
 
 test("validCondition", () => {
--- a/packages/stagebook/src/schemas/treatment.test.ts
+++ b/packages/stagebook/src/schemas/treatment.test.ts
@@ -79,6 +79,9 @@ test("reference timeline with no name", () => {
   if (!result.success)
     console.log(result.error.message, "\npath:", result.error.path);
   expect(result.success).toBe(false);
+  if (!result.success) {
+    expect(result.error.issues[0].message).toContain("A name must be provided");
+  }
 });
 
 // ----------- Condition Schema ------------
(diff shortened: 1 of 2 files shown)


JamesPHoughton:

---

PR #152, opened 2026-04-18
Align semantic tokens with raw source for quoted YAML scalars (#132)

### LINKED ISSUE #132: Semantic tokens: quoted YAML scalars produce misaligned ranges
## Problem

[semanticTokens.ts](apps/vscode/src/lib/semanticTokens.ts) computes token offsets and lengths from `value.range[0]` + `value.value`. For quoted YAML scalars the `range` from the `yaml` library typically includes the quote characters, while `value.value` does not. This misaligns the emitted semantic token by one character on each side whenever a user quotes a value, e.g.:

```yaml
type: "prompt"           # semantic token lands on `"prompt`, not `prompt`
file: "path/to/file.md"  # same shift
```

## Proposed approach

Derive token text/offsets from the raw source slice for the scalar node (`source.slice(range[0], range[1])`), then normalize quoting/escapes before emitting tokens. Add regression tests covering single-quoted, double-quoted, and unquoted scalars.

## Surfaced via

Copilot review on #130 (release-v0.4.0).

JamesPHoughton:
Closes #132.

## Problem

`computeSemanticTokens` in `apps/vscode/src/lib/semanticTokens.ts` emitted each token at `value.range[0]` with `value.value` as the text. For quoted YAML scalars the `yaml` library's `range` spans the surrounding quotes while `value.value` holds only the content, so every highlight was shifted one character to the left and one character short at the end:

```yaml
type: "prompt"           # token landed on `"prompt`, not `prompt`
file: "path/to/file.md"  # same shift
```

## Fix

Added a small `getScalarSource(range)` helper that reads `source.slice(range[0], range[1])` and, when the slice starts and ends with a matching `"` or `'`, strips the pair and bumps the offset by one. All call sites that previously used `(value.range[0], value.value)` now use `(scalarSrc.offset, scalarSrc.text)` — including the `file:` path splitter and the generic `${...}` template-variable emitter — so the on-screen token lines up with what the user typed regardless of quoting style. Set-membership checks (`elementTypeSet.has(...)`, etc.) still run against `value.value` so only semantically valid values are highlighted.

## Test plan

- [x] 8 new unit tests in `semanticTokens.test.ts` covering double-quoted and single-quoted variants of `type`, `file`, `reference`, `comparator`, plus template-var splitting inside a double-quoted file path and a double-quoted non-file value. All went red before the fix, green after.
- [x] `npm test` — stagebook 518, viewer 68, vscode 144 (↑8) — all pass.
- [x] `npm run lint` — zero warnings.
- [x] `npm run format:check` — clean.

### DIFF
--- a/apps/vscode/src/lib/semanticTokens.test.ts
+++ b/apps/vscode/src/lib/semanticTokens.test.ts
@@ -231,6 +231,118 @@ duration: 300`;
     });
   });
 
+  describe("quoted YAML scalars", () => {
+    it("aligns token for double-quoted element type", () => {
+      const src = `type: "prompt"`;
+      const tokens = computeSemanticTokens(src);
+      const typeTokens = tokens.filter((t) => t.tokenType === "type");
+      expect(typeTokens).toHaveLength(1);
+      expect(typeTokens[0]).toMatchObject({
+        line: 0,
+        startCol: 7,
+        length: 6,
+        text: "prompt",
+      });
+    });
+
+    it("aligns token for single-quoted element type", () => {
+      const src = `type: 'prompt'`;
+      const tokens = computeSemanticTokens(src);
+      const typeTokens = tokens.filter((t) => t.tokenType === "type");
+      expect(typeTokens).toHaveLength(1);
+      expect(typeTokens[0]).toMatchObject({
+        line: 0,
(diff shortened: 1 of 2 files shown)


JamesPHoughton:

---

PR #157, opened 2026-04-18
Add getReferencedAssets() utility and use it in VS Code asset check (#137)

### LINKED ISSUE #137: Export `getReferencedAssets()` utility for walking treatment trees and listing file references
## Context

Stagebook's element schemas know which fields on which element types point to files (e.g. `file`, `captionsFile`, `url` when it's a local path). Today this knowledge is implicit — consumers walk the tree and hard-code field names. The VS Code extension, for instance, checks only the literal field name `file` ([checkFileReferences in extension.ts](apps/vscode/src/extension.ts#L122-L203)), which means it silently under-validates:

- `mediaPlayer.captionsFile` — real local asset, not checked
- `mediaPlayer.url` — when it's a local path (not `http://…`), it's a file reference the schema accepts; not checked
- Any future element type that introduces a new file-like field name

Other platforms (e.g. the annotator that freezes treatments at activation time) have the same problem from the other direction: they need a list of every asset path a treatment touches so they can fetch and pin them, and they currently have to reverse-engineer that list from the schema.

The stagebook library is the right owner because it's the only place that authoritatively knows which element types have which file-like fields.

## Proposed utility

Add to `packages/stagebook/src/utils/referencedAssets.ts`:

```ts
export interface ReferencedAsset {
  /** The raw path as it appears in the treatment YAML. */
  path: string;
  /** Which field the path came from (e.g. "file", "captionsFile", "url"). */
  field: string;
  /** The element type ("prompt", "mediaPlayer", "image", "audio", …). */
  elementType: string;
  /** Element name if the element has one. */
  elementName?: string;
  /** Location of the value in the parsed object, useful for source mapping. */
  pathInTree: (string | number)[];
}

export function getReferencedAssets(treatmentFile: unknown): ReferencedAsset[];
```

## Which fields count as file references

Per-element-type allowlist (the utility owns this table, not each caller):

| Element type    | File-like fields                  |
|-----------------|-----------------------------------|
| `prompt`        | `file`                            |
| `image`         | `file`                            |
| `audio`         | `file`                            |
| `mediaPlayer`   | `url`, `captionsFile`             |
| `timeline`      | *(none — `source` is a name ref)* |

New element types with file fields must be added to this table; a small per-element-type test guards against drift.

## Exclusions (not returned)

- Entries whose `path` contains `${…}` (template placeholder) — resolution is a runtime concern
- Entries where the path is a full URL (`http://…`, `https://…`, `//…`) — the caller is interested only in local assets
- Paths that are empty strings

## Signature stability

The returned array's order is stable: tree order (outer-to-inner, treatment-then-introSequence-then-exit, array-index-order within each), then field-declaration-order within a single element. This lets consumers diff results meaningfully across treatment edits.

## Acceptance criteria

- Returns an entry for every file-like field on every element type listed above
- Excludes placeholder and URL paths per the rules above
- Tests cover each element type's file fields hitting both included and excluded paths
- Test covers correct `pathInTree` for an asset deep inside a templated gameStages array
- Test confirms full URL (`https://example.com/foo.mp4`) is excluded from `mediaPlayer.url` results
- Test confirms `${foo}/bar.mp4` is excluded even if surrounded by valid content
- Exported from `packages/stagebook/src/index.ts` as a public utility
- Used by `checkFileReferences` in the VS Code extension, replacing the hard-coded `file` check (follow-up — can be in the same PR or a subsequent one)

## Out of scope

- Path alias resolution (no `stagebook.config.json` awareness — that's its own separable concern in the remaining half of #26)
- Mutating or rewriting the treatment file
- Resolving paths to filesystem locations — this utility returns raw strings

## Split from

#26 — which will remain as the open design question about a config-file convention.

JamesPHoughton:
Closes #137.

## Summary

Stagebook's element schemas know which fields on which element types point to files. Until now that knowledge was implicit — consumers walked the tree and hard-coded field names. The VS Code extension's `checkFileReferences` only looked at literal `file`, so it silently missed `mediaPlayer.url` (when local) and `mediaPlayer.captionsFile`.

This PR adds `getReferencedAssets(treatmentFile: unknown): ReferencedAsset[]` as a public library utility that owns the per-element-type allowlist once, and rewires the VS Code extension to use it.

## The utility

`packages/stagebook/src/utils/referencedAssets.ts`:

```ts
export interface ReferencedAsset {
  path: string;
  field: string;
  elementType: string;
  elementName?: string;
  pathInTree: (string | number)[];
}

export function getReferencedAssets(treatmentFile: unknown): ReferencedAsset[];
```

Per-element-type allowlist (single source of truth):

| Element type  | File-like fields       |
|---------------|------------------------|
| `prompt`      | `file`                 |
| `image`       | `file`                 |
| `audio`       | `file`                 |
| `mediaPlayer` | `url`, `captionsFile`  |
| `timeline`    | *(none — name ref)*    |

Exclusions: template placeholders (`${…}`), full URLs (`http://…`, `https://…`, `//…`), empty strings. Accepts `unknown` so callers can feed raw YAML parse output *before* schema validation, as the downstream consumer examples in the issue do.

Order is stable tree-walk order (outer-to-inner, insertion order within each object), with field-declaration order per the table within a single element. This lets consumers diff results meaningfully across treatment edits.

## VS Code extension rewiring

`apps/vscode/src/extension.ts` `checkFileReferences` now iterates `getReferencedAssets(parsedObj)` instead of recursing for literal `file` keys. The path-traversal guard (`isWithinWorkspace`) is preserved. As a side effect, local `mediaPlayer.url` and `mediaPlayer.captionsFile` paths now get file-existence diagnostics too.

## Test plan

- [x] 21 unit tests in `referencedAssets.test.ts` covering each element type's file fields, every exclusion rule, deep `pathInTree` in templated `gameStages`, full-URL vs placeholder exclusion, and defensive `unknown` handling
- [x] `npm test` — stagebook 539, viewer 68, vscode 142 — all pass
- [x] `npm run lint` — clean (zero warnings)
- [x] `npm run build -w stagebook` — dts emits `getReferencedAssets` / `ReferencedAsset`
- [x] `npm run build -w stagebook-vscode` — extension bundles without error

https://claude.ai/code/session_012dNNMgyjojedBBx48fcADN

### DIFF
--- a/apps/vscode/src/extension.ts
+++ b/apps/vscode/src/extension.ts
@@ -13,7 +13,11 @@ import {
 import { expandAndValidate } from "./lib/expandAndValidate";
 import { findClosestMatch } from "./lib/levenshtein";
 import { isWithinWorkspace, relativizePath } from "./lib/filePaths";
-import { fillTemplates, treatmentFileSchema } from "stagebook";
+import {
+  fillTemplates,
+  getReferencedAssets,
+  treatmentFileSchema,
+} from "stagebook";
 import { parse as parseYaml } from "yaml";
 
 const diagnosticCollection =
@@ -127,8 +131,11 @@ function validateTreatmentFile(document: vscode.TextDocument): void {
 }
 
 /**
- * Walk the parsed treatment object looking for `file:` fields.
- * Check that referenced files exist and have the right extension.
+ * Walk the parsed treatment object and check that every local-asset path
+ * referenced by an element exists. `getReferencedAssets` owns the per-element-
+ * type allowlist of file-like fields (prompt.file, image.file, audio.file,
+ * mediaPlayer.url, mediaPlayer.captionsFile) and filters out template
+ * placeholders and full URLs; we still apply a path-traversal guard here.
  */
 function checkFileReferences(
   document: vscode.TextDocument,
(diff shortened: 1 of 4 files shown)


JamesPHoughton:

Copilot review addressed in 2f09b3d:
- `pathInTree` now points at the scalar value directly (JSDoc-aligned); VS Code extension no longer appends `field`
- Prompt shorthand supported (bare `*.prompt.md` strings inside `elements` arrays), with regression test
- `Object.hasOwn` guard so malformed YAML like `type: toString` can't turn the field-lookup into a non-array

CI green on 2f09b3d. Stagebook 546 / viewer 68 / vscode 142 tests passing.

---

Issue #250, opened 2026-04-30
Element types: remove `talkMeter` and `sharedNotepad`; deprecate `survey`


JamesPHoughton:
## Context

The Stagebook DSL has 14 element types. The wider syntax review (item R) flagged some of these as overlapping or redundant. After working through it, the answer is more nuanced than "consolidate" — most pairs that *look* redundant actually encode different intents and stay; a few are genuinely redundant or vestigial and go.

This is item #9 (the last) of the prioritized syntax review.

## The typology — what stays and why

The element types that *look* potentially redundant but **stay** because they encode distinct intents, not overlapping ones:

| Type          | Intent                                                                                       |
|---------------|----------------------------------------------------------------------------------------------|
| `audio`       | Non-interactive audio (chimes, alerts, ambient sound). One-shot; no participant controls expected. |
| `image`       | Non-interactive image display. Static picture; conceptually parallel to `audio`, not `display`. |
| `mediaPlayer` | Interactive audio/video with playback controls (play, pause, scrub, speed). Participant interacts with the clip. |
| `display`     | Shows a value looked up by reference from elsewhere in the study (e.g., `prompt.foo`'s response). Reactive to stored data, not a static resource. |

The discrimination matters: `audio` ≠ `mediaPlayer with playVideo:false` because `audio` is "fire and forget" while `mediaPlayer` is "give the participant a controllable surface." `image` ≠ `display` because `image` shows a static asset (URL or path) while `display` shows the value of a referenced data point. Each type answers a different authoring question.

These four are documented in the issue body above so the rationale is captured if anyone re-proposes consolidation.

### `talkMeter` — removed entirely

A platform-rendered indicator for who's speaking during a discussion. Niche, not consistently implemented across hosts, and discussions already manage their own speaker indication. Removing the element type and the corresponding `renderTalkMeter` field on `StagebookContext`.

### `sharedNotepad` — removed entirely

Redundant with `prompt` elements that have `type: openResponse` in their prompt file and `shared: true` set on the element. The shared-prompt path already routes through the host's `renderSharedNotepad` ([Prompt.tsx:161-167](packages/stagebook/src/components/elements/Prompt.tsx#L161-L167)); the standalone `sharedNotepad` element type is a parallel path that does the same thing.

Migration: replace each `sharedNotepad` element with a `prompt` element pointing at an openResponse prompt file with `shared: true`:

```yaml
# Before
- type: sharedNotepad
  name: groupNotes

# After
- type: prompt
[...]

Issue #249, opened 2026-04-30
Resources: separate `url:` (browser-direct) from `file:` (platform-resolved) with proper schemas


JamesPHoughton:
## Context

The schema today uses three different field names — `file:`, `url:`, and `captionsFile:` — for what's conceptually two distinct things, with inconsistent validation across them. This is item #8 of the wider syntax review.

### Inventory

| Element       | Field           | Schema today              |
|---------------|-----------------|---------------------------|
| `prompt`      | `file:`         | `promptFilePathSchema` (must end `.prompt.md`) |
| `audio`       | `file:`         | `fileSchema` (bare `z.string().optional()`) |
| `image`       | `file:`         | `fileSchema` (bare) |
| `mediaPlayer` | `url:`          | `z.string()` — **bare, no validation** |
| `mediaPlayer` | `captionsFile:` | `z.string()` — bare, no validation |
| `qualtrics`   | `url:`          | `urlSchema` (`https?://` or `asset://`) |
| `trackedLink` | `url:`          | `urlSchema` (`https?://` or `asset://`) |

Plus a footgun: [`elementBaseSchema`](packages/stagebook/src/schemas/treatment.ts#L871) declares `file: fileSchema.or(fieldPlaceholderSchema).optional()` at the base level, meaning **every element type silently accepts a `file:` key** even when it's meaningless (`separator`, `timer`, `submitButton`, etc.). The per-type `.extend(...).strict()` chain doesn't catch this — base fields pass through.

## The conceptual distinction

The current confusion is two different things sharing field names. Splitting them makes the schema reflect what's actually happening:

| | Resource the **browser** loads directly | Resource the **platform** helps resolve |
|--|--|--|
| Mental model | "Send the participant here" / "load this URL" | "Find me this thing" |
| Required form | Strictly a real `https?://` URL the browser can fetch | A handle the system resolves: relative path → resolved against treatment dir; `asset://` → host's `getAssetURL()`; `https?://` → fine, used directly |
| Field name | `url:` | `file:` |
| Elements | `trackedLink`, `qualtrics` | `prompt`, `audio`, `image`, `mediaPlayer`, `mediaPlayer.captionsFile` |

`url:` always means "what the browser uses." `file:` always means "what the platform looks up." That's consistent with how researchers actually think about the difference.

### Two distinct schemas

[...]

Issue #248, opened 2026-04-30
Docs: explain time-field reference frames (stage time vs media time vs durations)


JamesPHoughton:
## Context

Stagebook's time-valued fields use seconds as the unit but don't all share a reference frame. [Today's docs](docs/researcher/elements.md) cover each field individually but never spell out the mental model — researchers have to infer from which element type a field appears on. Worth one paragraph and a worked example.

The syntax review (item P in the original audit) initially flagged this as a structural concern. On review, the element type already provides enough context to disambiguate (a mediaPlayer-specific field is necessarily media-relative), and there's no clean rename that improves things without making the schema noisier. Docs treatment is the right scope.

## What needs to be added

A "Time fields and reference frames" section in [docs/researcher/elements.md](docs/researcher/elements.md) (or sibling) covering:

1. **Stage-relative time** — measured in seconds from when the stage started.
   - `displayTime` / `hideTime` (any element)
   - `timer.startTime` / `timer.endTime`
2. **Media-relative time** — measured in seconds from the start of the media file.
   - `mediaPlayer.startAt` / `mediaPlayer.stopAt`
3. **Durations** (interval lengths, no reference frame).
   - `stage.duration`
   - `mediaPlayer.stepDuration`
   - `timer.warnTimeRemaining`
4. **The bridge: `syncToStageTime`.** When set, mediaPlayer playback time is locked to stage elapsed time — the two frames coincide. This is worth surfacing in the section because it's the mechanism that makes "media-relative" and "stage-relative" relate to each other; it implicitly teaches the reader that the two frames exist.

## Worked example

```yaml
- type: mediaPlayer
  url: asset://clip.mp4
  displayTime: 5         # stage time: appear 5s into the stage
  hideTime: 60           # stage time: hide 60s into the stage
  startAt: 30            # media time: begin playback 30s into the clip
  stopAt: 90             # media time: pause 90s into the clip
```

A reader of this element should be able to say: "appears 5s after the stage begins, hides 55s later. While visible, the clip plays from 30s to 90s of media (so 60s of clip playback in 55s of stage time only works if speed is adjusted, otherwise the element hides before stopAt is reached)."

## Out of scope

- **Schema renames or restructuring.** Considered (`displayAt` instead of `displayTime`, `clipStartAt` instead of `startAt`, namespace blocks like `stage: { displayAt: ... }`); none was a clear win over the cost. The element type as discriminator already separates the frames at the type level.
[...]

Issue #246, opened 2026-04-30
References: rename `urlParams` source to `entryUrl.params.*` (resolve element-field collision)


JamesPHoughton:
## Context

The word `urlParams` does double duty in the schema today, naming two unrelated concepts:

1. **Element field** [`urlParams:`](packages/stagebook/src/schemas/treatment.ts#L1068-L1074) on `trackedLink` and `qualtrics` — an array of `{key, value | reference}` objects appended as query parameters to the **outgoing** URL when the participant clicks through or the iframe loads.
2. **Reference source** [`urlParams.<key>`](packages/stagebook/src/schemas/treatment.ts#L519-L532) — reads the **incoming** query params the participant arrived with.

Same word, opposite directions. The collision gets sharp when both appear in the same element:

```yaml
- type: trackedLink
  url: https://example.com
  urlParams:                          # outgoing (appending)
    - key: source
      reference: urlParams.referrer   # incoming (reading)
```

Two adjacent `urlParams` references with completely different semantics, no syntactic signal that they're different.

This is item #7 of the wider syntax review.

## Why this naming, not `entryParams` or `setUrlParams`/`getUrlParams`

Both fields really *are* URL parameters — that's the right noun. The actual difference isn't *what* they are, it's *which URL* they refer to:

- The element field's URL is right there on the element (`url: https://...`); `urlParams:` next to it unambiguously means "params for *this* URL." No ambiguity to resolve.
- The reference source has no visible URL — it's an implicit "the URL the participant arrived with." That's the URL that needs naming.

So the disambiguation happens by naming *which URL*, not by relabeling the data:

- Reference source: `urlParams.<key>` → `entryUrl.params.<key>`
- Element field: `urlParams:` (unchanged — still means "params for this element's URL")

Reads cleanly when both appear together:

```yaml
- type: trackedLink
  url: https://example.com
  urlParams:
    - key: source
      reference: entryUrl.params.referrer
```

The two URLs in this element are now obviously different things — one's the trackedLink's URL, one's the participant's entry URL.

Earlier framings that were considered and rejected:

- **`entryParams`** — drops the URL noun. Implies entry params could come from anywhere (form fields, deep links, embed parameters), losing precision about what kind of data this actually is.
- **`getUrlParams` / `setUrlParams`** — preserves "URL params" but verbs in YAML keys read off-register; everything else in the schema uses noun phrases.

Naming by URL identity preserves the technical accuracy of "URL params" while making the disambiguation explicit.

[...]

PR #251, opened 2026-04-30
Docs: visibility-field semantics + design-principles seed (closes #247)

### LINKED ISSUE #247: Docs: spell out visibility-field semantics (AND combination, intro/exit constraints, time reference frame)
## Context

The element visibility fields (`displayTime`, `hideTime`, `showToPositions`, `hideFromPositions`, `conditions`) each have a one-line description in [docs/researcher/elements.md](docs/researcher/elements.md#L8-L11), but the cross-field semantics aren't written down anywhere researchers will find them. The syntax review (item D in the original audit) initially flagged this as a structural problem, but on review the fields are doing two clear, complementary jobs (allow-list and deny-list) that combine in the standard ACL pattern. The fix is documentation, not a schema break.

## What needs to be added

A "Visibility" section in the element reference (or wherever fits cleanly) that covers:

1. **All visibility fields combine with implicit AND.** An element is visible only when *every* visibility field that's set evaluates to "show." Specifically:
   - Current stage time is at or after `displayTime` (if set) and before `hideTime` (if set).
   - Participant's position is in `showToPositions` (if set) and not in `hideFromPositions` (if set).
   - All `conditions` evaluate to true (if any).
2. **`displayTime` and `hideTime` are stage-relative seconds**, measured from when the stage started. The reference frame isn't in the field name; it's worth saying so.
3. **`showToPositions` and `hideFromPositions` aren't allowed in intro/exit steps.** Intro/exit steps run per participant, before they have a position assignment — the schema rejects these fields there. Worth telling researchers so they don't write them and get cryptic preflight errors.
4. **Worked example combining several visibility rules** so the AND semantic is concrete:
   ```yaml
   - type: prompt
     file: prompts/post_position.prompt.md
     displayTime: 30        # appears after 30s
     showToPositions: [0]   # only player 0 sees it
     conditions:
       - reference: prompt.consent
         comparator: equals
         value: yes          # and only if they consented
   ```

## Out of scope

- **Schema changes.** Keeping the five fields as-is — the original "additive masking is awkward" complaint doesn't survive a closer look. The two-field shape (allow-list + deny-list) communicates two different intents clearly and matches the standard ACL pattern.
- **Folding visibility into `conditions:`.** Considered and deferred (see syntax-review thread for the cost-benefit analysis). Time-as-data and position-as-data would meaningfully expand the condition evaluator's surface area; not worth the trade today.

## Acceptance

- [ ] [docs/researcher/elements.md](docs/researcher/elements.md) (or a sibling) gains a "Visibility" section covering the four points above.
- [ ] At least one worked example combines time, position, and condition visibility in the same element.
- [ ] Cross-link from [docs/researcher/conditions.md](docs/researcher/conditions.md) to the new section, since `conditions:` is one of the AND'd visibility mechanisms.

## Related

- Parent syntax review (item D in the wider DSL audit; reframed as docs-only after closer analysis).

JamesPHoughton:
## Summary

- Adds a **Visibility** section to `docs/researcher/elements.md` documenting the implicit-AND combination of the five visibility fields (`displayTime`, `hideTime`, `showToPositions`, `hideFromPositions`, `conditions`), the stage-relative time reference frame, the intro/exit-step constraint on positions, and a worked example combining all four mechanisms.
- Cross-links from `docs/researcher/conditions.md` to the new section so researchers reading conditions docs see the AND-with-other-visibility-fields interaction.
- Seeds `docs/decisions/principles.md`, the synthesis document for the nine cross-cutting design principles surfaced in the wider syntax review. Subsequent syntax-review issues (#235, #238, #240, #243, #244, #245, #246, #248, #249, #250) link into specific principles; landing the seed here keeps those references resolvable.

## Test plan

- [x] Prettier passes (`npx prettier --check`) on all three files.
- [x] No code changes — schemas, runtime, and components untouched.
- [x] Pre-push hook passed (lint + tests across workspaces).
- [ ] CI green on the PR.
- [ ] Manual review of the rendered Markdown in the GitHub PR preview to confirm the new section reads cleanly and the cross-links resolve.


JamesPHoughton:
Addressed both Copilot review comments in caa5b48:

1. **elements.md intro/exit-step constraint** — the doc claimed the schema rejects `showToPositions`/`hideFromPositions` in both intro and exit steps, but only `introStepsSchema`'s superRefine actually has those rules; `exitStepsSchema` doesn't. Corrected the doc to scope the restriction to intro steps and added a sentence clarifying that exit steps run after position assignment so position-based visibility works there the same as in game stages. This matches the actual schema behavior rather than narrowing it as the suggestion would have done.

2. **principles.md grammar nit** — applied the suggestion verbatim (removed the duplicate "this").

Re-requesting Copilot review.

PR #252, opened 2026-04-30
Docs: time-field reference frames (closes #248)

### LINKED ISSUE #248: Docs: explain time-field reference frames (stage time vs media time vs durations)

JamesPHoughton:
## Summary

Adds a **Time fields and reference frames** section to `docs/researcher/elements.md` covering:

- **Stage-relative time** — `displayTime`, `hideTime`, `timer.startTime`, `timer.endTime` (measured from when the stage started).
- **Media-relative time** — `mediaPlayer.startAt`, `mediaPlayer.stopAt` (measured from the beginning of the media file).
- **Durations (no frame)** — `stage.duration`, `mediaPlayer.stepDuration`, `timer.warnTimeRemaining` (interval lengths, not points in time).
- **Bridging the two frames** — `syncToStageTime` cross-referenced as the mechanism that locks media playback time to stage elapsed time when the frames need to coincide.
- **Worked example** combining stage-frame and media-frame fields on the same `mediaPlayer` element, including the failure case where the visibility window closes before media `stopAt` is reached.

Closes #248.

## Test plan

- [x] Prettier passes (`npx prettier --check`).
- [x] No code changes — schemas, runtime, and components untouched.
- [x] Pre-push hook passed (lint + tests across workspaces).
- [ ] CI green on the PR.
- [ ] Copilot review addressed; re-review confirmed via @copilot mention (testing this re-review approach).

### DIFF
--- a/docs/researcher/elements.md
+++ b/docs/researcher/elements.md
@@ -522,6 +522,59 @@ Displays cumulative speaking time for each participant. Only meaningful in video
 - type: talkMeter
 ```
 
+## Time fields and reference frames
+
+Several element fields are measured in seconds, but they don't all share the same reference frame. Knowing which frame a field uses matters when you mix them — most often when a `mediaPlayer` is timed against the stage clock.
+
+### Stage-relative time
+
+Measured in seconds from when the **stage** started.
+
+| Field             | Element     | Says                                          |
+| ----------------- | ----------- | --------------------------------------------- |
+| `displayTime`     | any element | Show after this many seconds into the stage   |
+| `hideTime`        | any element | Hide after this many seconds into the stage   |
+| `timer.startTime` | `timer`     | When the timer's progress bar starts counting |
+| `timer.endTime`   | `timer`     | When the timer's progress bar reaches 100%    |
+
+### Media-relative time
+
+Measured in seconds from the beginning of the **media file**.
+
+| Field                 | Element       | Says                                       |
+| --------------------- | ------------- | ------------------------------------------ |
+| `mediaPlayer.startAt` | `mediaPlayer` | Position in the clip where playback begins |
+| `mediaPlayer.stopAt`  | `mediaPlayer` | Position in the clip where playback pauses |
+
+### Durations (no frame)
+
+Lengths of intervals, not points in time. They have no reference frame.
+
+| Field                      | Element       | Says                                        |
+| -------------------------- | ------------- | ------------------------------------------- |
+| `stage.duration`           | stage         | How long the stage lasts                    |
+| `mediaPlayer.stepDuration` | `mediaPlayer` | How much each "step" key advances           |
+| `timer.warnTimeRemaining`  | `timer`       | Seconds-from-end at which the bar turns red |
+
+### Bridging the two frames: `syncToStageTime`
+
+When `mediaPlayer.syncToStageTime: true` is set, the player's playback position is locked to stage elapsed time — the two frames coincide. With sync on, the player's controls are not used (they're forbidden by the schema), and a stage second always corresponds to a media second.
+
+### Combining frames on one element
+
+A `mediaPlayer` element using both frames at once:
+
+```yaml
+- type: mediaPlayer
+  url: https://example.com/clip.mp4
+  displayTime: 5 # stage time: appear 5s into the stage
+  hideTime: 60 # stage time: hide 55s later
+  startAt: 30 # media time: begin playback 30s into the clip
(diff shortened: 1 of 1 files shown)


JamesPHoughton:
@copilot — addressed all six review comments in b19a1a6:

1. **Duplication with the Visibility section's "Time reference frame" subsection** (your line 539 comment) — the Visibility section now has a one-sentence pointer to the new canonical section instead of partial coverage. No drift risk.
2. **`timer.startTime` / `timer.endTime` notation** (line 538) — dropped the `timer.` prefix; the table lists bare field names with the Element column saying where each lives. Added an explanatory note.
3. **`mediaPlayer.startAt` / `mediaPlayer.stopAt` notation** (line 547) — same fix; bare `startAt` / `stopAt` with the Element column.
4. **`mediaPlayer.stepDuration` / `timer.warnTimeRemaining` notation** (line 557) — same fix; bare `stepDuration` / `warnTimeRemaining`.
5. **`syncToStageTime` description over-strong** (line 561) — rewrote to match the implementation: it's a one-time seek on mount (`startAt + currentStageElapsedTime`), not continuous locking. Calls out drift on pause/buffer and the schema's `controls` exclusion.
6. **Worked example "lock the two clocks together" misleading** (line 576) — replaced with a concrete walk-through: with `syncToStageTime: true`, the clip mounts at media-time 35 and plays through to 90 right as the element hides at stage-time 60, conditional on uninterrupted playback.

Could you take another look?

JamesPHoughton:
@copilot-pull-request-reviewer — addressed all six review comments in b19a1a6. Could you re-review and confirm whether anything else stands out?

PR #253, opened 2026-04-30
feat(conditions): boolean-tree operators all/any/none (closes #235)

### LINKED ISSUE #235: Conditions: nested boolean logic with `all` / `any` / `none` operators
## Context

Today, every \`conditions:\` block in the treatment DSL is a flat array of leaf conditions, AND'd together. See [conditionsSchema in treatment.ts](packages/stagebook/src/schemas/treatment.ts#L837-L846):

```yaml
conditions:
  - reference: prompt.foo
    comparator: equals
    value: yes
  - reference: prompt.bar
    comparator: exists
```

There's no way to express OR, NOR, or grouped boolean logic. Researchers either:

- duplicate the same stage with different conditions to fake an OR, or
- contort the predicate using the negated-comparator twins (`doesNotEqual`, `doesNotExist`, `isNotOneOf`, …) when De Morgan equivalents happen to fit, or
- give up on the feature.

Bolting nested logic onto the array shape after the fact is very disruptive — every consumer of the DSL (the cross-stage walker in [validateReferences.ts](packages/stagebook/src/schemas/validateReferences.ts), the runtime evaluator in [evaluateConditions.ts](packages/stagebook/src/utils/evaluateConditions.ts), the preflight rule walker [validateConditionRules](packages/stagebook/src/schemas/treatment.ts#L784-L835), the viewer's UI, third-party tooling) assumes the flat shape. Doing it during the current syntax refactor avoids that pain.

This is item #1 in the syntax review (parent discussion forthcoming).

## Motivation — cases that hit the limit today

- **OR across references.** Show a follow-up element if either participant's previous answer was "yes" *or* the timer ran out. Currently inexpressible without duplicate stages.
- **NOR / fallback.** Render a "thanks, your group is complete" message if *none* of the players answered the slider above 50. Currently requires an awkward chain of negated-comparator leaves and gives wrong results when the underlying value is undefined.
- **Mixed depth.** "Skip this stage if (P1 disagrees AND P2 disagrees) OR (everyone abstained)." Today this needs a fan-out into multiple stages with a shared exit step.

## Proposal

Operator-keyed object nodes — `all:` / `any:` / `none:` — recursive, with leaf conditions (the existing `{reference, comparator, value, position}` object) as terminals. Plus a single root-level sugar: an array at the top of a `conditions:` field is implicit `all:`, so every existing treatment file keeps working unchanged.

### Vocabulary

- `all: [A, B, …]` — true when every child is true (AND)
- `any: [A, B, …]` — true when at least one child is true (OR)
- `none: [A, B, …]` — true when zero children are true (NOR)

`all` / `any` / `none` is a symmetric trio that reads as plain English and matches the existing `position: all | any` aggregator vocabulary at the player level (which has the same semantics applied across players, not across conditions).

We deliberately do *not* introduce a unary `not:`. A unary negation is just `none: [X]`, and `not:` followed by a list (`not: [A, B]`) would be ambiguous between "not (A and B)" (NAND) and "not (A or B)" (NOR). Forcing the list-shaped operator settles the ambiguity.

### Examples — before / after

**Existing flat AND (unchanged, no migration):**

```yaml
conditions:
  - reference: prompt.foo
    comparator: equals
    value: yes
  - reference: prompt.bar
    comparator: exists
```

**Explicit `all:` form (equivalent):**

```yaml
conditions:
  all:
    - { reference: prompt.foo, comparator: equals, value: yes }
    - { reference: prompt.bar, comparator: exists }
```

**OR — show element if either participant changed their mind:**

```yaml
conditions:
  any:
    - { reference: prompt.remote_work_changed, position: 0, comparator: exists }
    - { reference: prompt.remote_work_changed, position: 1, comparator: exists }
```

**NOR — fallback message when nobody hit the threshold:**

```yaml
conditions:
  none:
    - { reference: prompt.familiarity, position: 0, comparator: isAtLeast, value: 50 }
    - { reference: prompt.familiarity, position: 1, comparator: isAtLeast, value: 50 }
```

**Mixed depth — (P1 OR P2 disagrees) AND timer not exceeded:**

```yaml
conditions:
  all:
    - any:
        - reference: prompt.consensus
          position: 0
          comparator: equals
          value: disagree
        - reference: prompt.consensus
          position: 1
          comparator: equals
          value: disagree
    - reference: prompt.discussion_overflow
      comparator: doesNotExist
```

## Schema sketch

```ts
const leafConditionSchema = /* the existing discriminated-on-comparator
                                reference/comparator/value/position object */;

const conditionNodeSchema: z.ZodType = z.lazy(() =>
  z.union([
    z.object({ all:  z.array(conditionNodeSchema).nonempty() }).strict(),
    z.object({ any:  z.array(conditionNodeSchema).nonempty() }).strict(),
    z.object({ none: z.array(conditionNodeSchema).nonempty() }).strict(),
    leafConditionSchema,
  ]),
);

// Field-level: array at root is sugar for all:
export const conditionsSchema = z.union([
  z.array(conditionNodeSchema).nonempty(),
  conditionNodeSchema,
]);
```

Notes:

- `nonempty()` everywhere — `all: []` would be vacuously true and `any: []` vacuously false; both are author-error shapes worth rejecting at preflight.
- Operator objects are `.strict()` so combinations like `{all: [...], reference: ...}` fail loudly.
- Array-as-AND sugar is **root-only**. Inside `all:` / `any:` / `none:`, every child is a single node. No nested array shorthand — it would muddy "is this list a sub-AND or just the operator's children".
- Pre-scan operator keys with a `superRefine` to catch typos (`al:` → "did you mean `all`?") before the union runs, since zod's union errors are unhelpful otherwise.
- `altTemplateContext` (template-fill compatibility) wraps the field-level schema as today, so `conditions: { template: someTemplateName }` continues to expand. Open question: do we also allow templates *inside* an operator branch (e.g. `all: [{ template: ... }, ...]`)? See open questions below.

## Evaluator semantics (tri-state)

[`compare()` already returns `true | false | undefined`](packages/stagebook/src/utils/compare.ts), where `undefined` means "the referenced data isn't there yet — can't decide." Combinators preserve that:

- **`all`**: any child false → false; else any undefined → undefined; else true.
- **`any`**: any child true → true; else any undefined → undefined; else false.
- **`none`**: any child true → false; else any undefined → undefined; else true.

This matches Kleene three-valued logic and is what users would intuit. The callsites in [`evaluateConditions.ts`](packages/stagebook/src/utils/evaluateConditions.ts) collapse `undefined` to the appropriate fallback at the boundary (typically: skip-at-load uses `false`; mid-stage early-end is more conservative).

## Backward compatibility

- Every existing treatment file is valid as-is. The flat-array form is sugar for `all:` and stays the canonical short-form.
- Existing tests that assert flat-array shapes keep passing.
- The negated comparator twins (`doesNotEqual`, `doesNotExist`, `doesNotInclude`, `doesNotMatch`, `isNotOneOf`, plus the symmetric numeric pairs) are **kept**, not deprecated. Reason: `none:` expresses NOR but not NAND, so the negated leaves remain the only way to write "not (A and B)" without nested gymnastics. Deduplicating the comparator list would be a separate breaking change requiring a full unary `not:` operator, and isn't worth the trade.

## Out of scope / future work

- **Counting operators** — `atLeast: { n: 3, of: [...] }`, `atMost`, `exactly`. These strictly subsume `all` / `any` / `none` (`all = atLeast(n)`, `any = atLeast(1)`, `none = atMost(0)`) but cover the rarer "k of n" case, e.g., "answered 3 of these 5 attention checks correctly." Adding them later is purely additive in the union — no breakage. Defer to a follow-up.
- **Generalized cross-player aggregation.** Today `condition.position` overloads "which player's data" with "how to aggregate" (`all`, `any`, `percentAgreement`, slot index). The boolean-tree work here is independent of that overload — they're about combining *across conditions*, not *across players*. The position-aggregator cleanup is its own issue (item #2 of the syntax review).
- **Removing negated-comparator twins.** As above — keeping them buys NAND expressiveness, removing them would require a unary `not:` operator we've decided against. Out of scope for this issue.

## Open questions

1. **Templates inside operator branches.** Today `conditions: { template: foo }` expands at preflight. Should `any: [{ template: foo }, ...]` also expand, or should we restrict templates to wrap whole condition trees? Permissive is consistent with current `altTemplateContext` behavior; restrictive simplifies the schema. *Lean: permissive for consistency.*
2. **Should preflight reject double negation?** `none: [{ none: [X] }]` is `X` — author confusion likely. *Lean: warn, don't reject — it's not wrong, just unidiomatic, and a researcher building up a tree iteratively might hit it.*
3. **Should `validateConditionRules` (game-stage `forbidSelfPosition`, `percentAgreement` numeric-comparator) recurse into operator branches?** Yes — same constraints apply at every depth. The walker just needs to descend into `all`/`any`/`none` children. Worth calling out so the implementation PR remembers it.

## Acceptance

- [ ] `conditionsSchema` accepts the existing flat-array form unchanged.
- [ ] `conditionsSchema` accepts `all:` / `any:` / `none:` operator nodes at any depth.
- [ ] Operator nodes reject extra keys (`.strict()`).
- [ ] Empty operator arrays rejected with a clear message.
- [ ] Pre-scan typo'd operator keys with a friendly suggestion.
- [ ] [`evaluateConditions.ts`](packages/stagebook/src/utils/evaluateConditions.ts) handles operator nodes with tri-state semantics; existing leaf evaluation unchanged.
- [ ] [`validateTreatmentFileReferences`](packages/stagebook/src/schemas/validateReferences.ts) recurses into operator branches when collecting reference sites.
- [ ] [`validateConditionRules`](packages/stagebook/src/schemas/treatment.ts#L784-L835) recurses into operator branches when applying game-stage and percentAgreement rules.
- [ ] Unit tests cover: flat-array compat, single-operator at root, nested, empty-array rejection, typo detection, tri-state evaluation per operator.
- [ ] Documentation example added to the annotated walkthrough showing one OR or NOR usage.

## Related

- #183 — stage-level conditions (introduced the desync constraint that recurses cleanly under operators)
- #197 — cross-stage reference walker (needs to recurse into operator branches)
- #199 — stage-advance telemetry (advance reasons want to identify which sub-branch of a tree fired)
- #126 — `exists` on unanswered references (tri-state evaluator semantics interact here)

JamesPHoughton:
## Summary

Adds nested boolean logic to `conditions:` via three operator-keyed nodes — `all: [...]`, `any: [...]`, `none: [...]` — that recurse, with leaf conditions at the bottom of the tree. Backward compatible: a top-level array stays valid as sugar for `all: [...]`, so every existing treatment file continues to validate unchanged.

The pivotal correctness case is the tri-state evaluator's handling of `none:` over unknown leaves — without three-valued logic, fallback elements gated on `none:` would render before any participant had answered. The implementation propagates `undefined` through Kleene three-valued logic and collapses to `false` at the public boundary, so fallbacks stay hidden until at least one answer arrives.

This PR was developed under the local-convergence review protocol: implement → run tests + lint → independent review subagent (3 passes) → resolve substantive findings → push. The subagent caught one substantive bug on the second pass (Rule 2 false-positive for leaves nested inside `any:`/`none:` operators) which is fixed and tested. The third pass found no substantive issues.

## What's in the PR

- **Schema** (`treatment.ts`, `resolved.ts`): recursive `conditionNodeSchema` (`z.lazy` union with `.strict().nonempty()` operator branches), `conditionsSchema` accepting array (sugar) or single node, operator-key typo detection (`al` → suggest `all`), `OPERATOR_KEYS` exported as the source of truth.
- **Tri-state evaluator** (`evaluateConditions.ts`): internal `evaluateNode` returns `true | false | undefined`, operators propagate per Kleene three-valued logic, public boundary collapses unknown → false.
- **Walkers** (`validateReferences.ts`, `treatment.ts`): cross-stage walker uses `walkConditionLeaves` generator; per-rule walker recurses through operators with correct path scoping. Rule 2 (always-skip-at-load) gated to leaves whose path doesn't traverse `any:`/`none:` (where per-leaf simulation isn't sound).
- **Components** (`StageConditionGate.tsx`, `ConditionsConditionalRender.tsx`): widened to accept arrays, operator nodes, and leaves; `ConditionNode` re-exported through both component barrels.
- **Docs** (`docs/researcher/conditions.md`): "Boolean operators" section with OR/NOR examples, nested usage, and a "Three-valued logic" subsection explaining what happens before data arrives.

## Test plan

- [x] All 733 vitest tests pass (688 baseline + ~45 new across `evaluateConditions.test.ts`, `treatment.test.ts`, `validateReferences.test.ts`).
- [x] Lint clean (`eslint src/`).
- [x] Prettier clean across the diff.
- [x] Backward compat: existing flat-array conditions validate and evaluate unchanged.
- [x] Pivotal `none: [allUnknown] → undefined → false` case covered with a dedicated test.
- [x] Walker recursion verified via `forbidSelfPosition` reaching a leaf nested in `any:`.
- [x] Rule 2 path-awareness covered: positive cases for flat-array and `all:`, negative cases for `any:`, `none:`, and nested-via-`any:`.
- [ ] CI green on the PR.
- [ ] Copilot first-pass review (acting as backstop; local convergence already ran via 3 review-subagent passes).

### DIFF
--- a/docs/researcher/conditions.md
+++ b/docs/researcher/conditions.md
@@ -23,7 +23,99 @@ conditions:
     value: 15
 ```
 
-For OR logic, create multiple elements with different conditions.
+## Boolean operators: `all`, `any`, `none`
+
+When you need OR or NOR logic, wrap conditions in an operator-keyed object. The flat-array form above is sugar for `all:` — these two are equivalent:
+
+```yaml
+# Implicit all (sugar)
+conditions:
+  - reference: prompt.a
+    comparator: equals
+    value: yes
+  - reference: prompt.b
+    comparator: exists
+
+# Explicit all
+conditions:
+  all:
+    - { reference: prompt.a, comparator: equals, value: yes }
+    - { reference: prompt.b, comparator: exists }
+```
+
+Use `any:` for OR, `none:` for NOR (none of these are true):
+
+```yaml
+# Show element if either participant's previous answer was "yes"
+conditions:
+  any:
+    - {
+        reference: prompt.changedMind,
+        position: 0,
+        comparator: equals,
+        value: yes,
+      }
+    - {
+        reference: prompt.changedMind,
+        position: 1,
+        comparator: equals,
+        value: yes,
+      }
+```
+
+```yaml
+# Render fallback message when nobody hit the threshold
+conditions:
+  none:
+    - {
+        reference: prompt.familiarity,
+        position: 0,
+        comparator: isAtLeast,
+        value: 50,
+      }
+    - {
+        reference: prompt.familiarity,
+        position: 1,
+        comparator: isAtLeast,
+        value: 50,
+      }
+```
+
+Operators nest. Mix freely:
+
+```yaml
+# (P1 disagrees OR P2 disagrees) AND timer hasn't overflowed
+conditions:
+  all:
+    - any:
+        - {
+            reference: prompt.consensus,
+            position: 0,
+            comparator: equals,
+            value: disagree,
+          }
+        - {
+            reference: prompt.consensus,
+            position: 1,
+            comparator: equals,
+            value: disagree,
+          }
+    - { reference: prompt.discussion_overflow, comparator: doesNotExist }
+```
+
+### Three-valued logic — what happens before data arrives
+
+Each leaf condition can be **true**, **false**, or **unknown** (data not yet recorded). Operators propagate "unknown" so fallback elements gated on `none:` don't render prematurely:
+
+| Operator | True when            | False when           |
+| -------- | -------------------- | -------------------- |
+| `all`    | every child is true  | any child is false   |
+| `any`    | any child is true    | every child is false |
+| `none`   | every child is false | any child is true    |
+
+If neither row applies (because some children are still "unknown"), the operator itself is unknown — at the rendering boundary, that collapses to "don't show yet." The most common place this matters: a `none:` block whose children all reference data nobody has answered yet stays hidden until at least one answer arrives, instead of rendering as if "no one matched."
+
+For OR logic on a single reference (across positions, comparators, etc.), `any:` is usually clearer than creating separate elements. For NOR, `none:` replaces the De Morgan trick of stacking negated comparators (`doesNotEqual`, `doesNotInclude`, `isNotOneOf`).
 
 **Visibility-field interaction.** When an element also has `displayTime`, `hideTime`, `showToPositions`, or `hideFromPositions` set, all of those fields combine with `conditions` using implicit AND — the element is visible only when every visibility field that's set evaluates to "show." See [Element visibility](elements.md#visibility) for the full picture.
 
--- a/packages/stagebook/src/components/conditions/ConditionsConditionalRender.tsx
+++ b/packages/stagebook/src/components/conditions/ConditionsConditionalRender.tsx
@@ -1,57 +1,41 @@
 import React from "react";
 import {
-  evaluateCondition,
+  evaluateConditions,
   type Condition,
+  type ConditionNode,
 } from "../../utils/evaluateConditions.js";
 
-export type { Condition };
+export type { Condition, ConditionNode };
 
 export interface ConditionsConditionalRenderProps {
-  conditions: Condition[];
+  /** A `conditions:` value: a flat array (implicit `all`), an operator
+   *  node (`{all|any|none: [...]}`), a single leaf condition, or
+   *  null/undefined (which renders children unconditionally). */
+  conditions: ConditionNode[] | ConditionNode | null | undefined;
   resolve: (reference: string, position?: string) => unknown[];
   children: React.ReactNode;
   fallback?: React.ReactNode;
 }
 
+/**
+ * Renders `children` when the condition tree evaluates to true,
+ * `fallback` otherwise. Delegates to `evaluateConditions`, which
+ * handles all three shapes (array sugar, operator node, leaf) and
+ * collapses the tri-state "data not yet" case to false.
+ */
 export function ConditionsConditionalRender({
   conditions,
   resolve,
   children,
   fallback = null,
 }: ConditionsConditionalRenderProps) {
-  if (!conditions || !conditions.length) return <>{children}</>;
-
-  return (
-    <RecursiveConditionalRender
-      conditions={conditions}
-      resolve={resolve}
-      fallback={fallback}
-    >
-      {children}
-    </RecursiveConditionalRender>
-  );
-}
-
-function RecursiveConditionalRender({
-  conditions,
-  resolve,
-  children,
-  fallback = null,
-}: ConditionsConditionalRenderProps) {
-  const condition = conditions[0];
-  const referenceValues = resolve(condition.reference, condition.position);
-  const conditionMet = evaluateCondition(condition, referenceValues);
-
-  if (!conditionMet) return <>{fallback}</>;
-  if (conditions.length === 1) return <>{children}</>;
-
-  return (
-    <RecursiveConditionalRender
-      conditions={conditions.slice(1)}
-      resolve={resolve}
-      fallback={fallback}
-    >
-      {children}
-    </RecursiveConditionalRender>
-  );
+  if (
+    conditions === undefined ||
+    conditions === null ||
+    (Array.isArray(conditions) && conditions.length === 0)
+  ) {
+    return <>{children}</>;
+  }
+  const conditionMet = evaluateConditions(conditions, resolve);
+  return <>{conditionMet ? children : fallback}</>;
 }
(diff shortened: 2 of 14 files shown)


JamesPHoughton:
Addressed two of three Copilot comments in 871aee0:

1. **`resolved.ts:81` (type weakness)** — Added a `ResolvedConditionNode` TS type and threaded it into `z.ZodType<ResolvedConditionNode>`. `z.infer<typeof resolvedConditionSchema>` now resolves to the structured union instead of `any`, so external consumers of `ResolvedConditionType` keep type safety.
2. **`validateReferences.ts:25` (circular dependency)** — Real issue: `treatment.ts` already imports `validateReferences.ts` for the cross-stage walker, so importing `OPERATOR_KEYS` back from `treatment.ts` formed a cycle. Extracted to a new third module (`packages/stagebook/src/schemas/conditionOperators.ts`); both files import from there, `treatment.ts` re-exports for the public API.

The third comment (markdown table double-pipe in `docs/researcher/conditions.md:114`) appears to be a misread of the diff — the file's table rows use standard single-pipe GFM (`| Operator | True when | False when |`). Confirmed locally with `grep "^||"` returning nothing. No change needed.

Tests still 733/733 passing; lint clean. Branch is ready for re-review.

PR #254, opened 2026-04-30
feat(conditions): narrow position to a read selector (closes #238)

### LINKED ISSUE #238: Conditions: narrow `position` to a read selector (drop `all`, `any`, `percentAgreement`)
## Context

Today, `position` on a condition leaf ([baseConditionSchema in treatment.ts](packages/stagebook/src/schemas/treatment.ts#L549-L552)) accepts `shared | player | all | any | percentAgreement | <slot index>`. That value space is doing three different jobs in one field:

| value               | role             |
|---------------------|------------------|
| `<N>`, `shared`     | read selector    |
| `player` (default)  | read selector    |
| `all`               | fan-out + AND    |
| `any`               | fan-out + OR     |
| `percentAgreement`  | numeric metric — compute consensus rate, compare against `value` |

The overload has caused real bugs already — [`validateConditionRules`](packages/stagebook/src/schemas/treatment.ts#L784-L835) exists in part because the default `player` value on a game-stage-level condition would desync players, and `percentAgreement` can only be combined with numeric comparators because the field's *meaning* silently shifts from "raw value" to "agreement rate" depending on which keyword is present.

This is item #2 of the wider syntax review. It pairs with #235 (nested boolean operators) — the boolean tree absorbs the aggregator cases, leaving `position` with only its read-selector job to do.

## Motivation

With `all:` / `any:` / `none:` operators landing in #235, the aggregator values on `position` become expressible at the tree level:

- `position: all` ≡ `all: [<one leaf per slot>]`
- `position: any` ≡ `any: [<one leaf per slot>]`

So they stop earning their keep — the same predicate is more uniformly written as a tree, and the asymmetry of "we sugar AND across players but not OR across players" goes away. `percentAgreement` is fundamentally different (numeric rate, not a boolean predicate); it's awkward to use, rare in practice, and worth dropping cleanly so a future *countables / aggregates* family can be designed deliberately rather than retrofitted into `position`.

Narrowing `position` to a pure read selector removes the overload, simplifies the schema, and makes the field's role one sentence to explain: **where to read the referenced value from.**

## Proposal

`position` on a condition leaf accepts only:

| value          | meaning                                      |
|----------------|----------------------------------------------|
| `<slot index>` | read player N's value                        |
| `shared`       | read the shared store                        |
| `player`       | read this participant's own value (default)  |

`all`, `any`, and `percentAgreement` are removed.

This is a hard breaking change. The `all` / `any` cases have a mechanical migration via #235's boolean tree; `percentAgreement` does not — see Migration below.

## Examples — before / after

**Before:**

```yaml
# AND across players
- reference: prompt.consent
  position: all
  comparator: equals
  value: yes

# OR across players
- reference: prompt.flag
  position: any
  comparator: exists

# Consensus rate
- reference: prompt.agree
  position: percentAgreement
  comparator: isAtLeast
  value: 60
```

**After** (using #235's boolean tree):

```yaml
# AND across players — explicit fan-out
all:
  - { reference: prompt.consent, position: 0, comparator: equals, value: yes }
  - { reference: prompt.consent, position: 1, comparator: equals, value: yes }

# OR across players — explicit fan-out
any:
  - { reference: prompt.flag, position: 0, comparator: exists }
  - { reference: prompt.flag, position: 1, comparator: exists }

# Consensus rate — no direct equivalent. See "Out of scope" below.
```

## Schema diff

[`baseConditionSchema.position`](packages/stagebook/src/schemas/treatment.ts#L549-L552) today:

```ts
position: z
  .enum(["shared", "player", "all", "any", "percentAgreement"])
  .or(z.number().nonnegative().int())
  .optional(),
```

After:

```ts
position: z
  .enum(["shared", "player"])
  .or(z.number().nonnegative().int())
  .optional(),
```

Cleanups that follow:

- Drop [`PERCENT_AGREEMENT_NUMERIC_COMPARATORS`](packages/stagebook/src/schemas/treatment.ts#L699-L704) and the `validateConditionRules` rule that requires numeric comparators when `position === "percentAgreement"`.
- Update [`resolvedConditionSchema.position`](packages/stagebook/src/schemas/resolved.ts#L33-L36) to match the narrowed enum.
- The game-stage `forbidSelfPosition` rule ([`GAME_STAGE_FORBIDDEN_POSITIONS`](packages/stagebook/src/schemas/treatment.ts#L710-L713)) is **unchanged** — `player` (or default) is still rejected at game-stage level for the same desync reason.
- [`positionSelectorSchema`](packages/stagebook/src/schemas/treatment.ts#L122-L126), used by `display.position` and `trackedLink` URL params, is **not** changed. `display.position: all` is a render concern (show all players' values), not a condition aggregator. Different field, different overload — out of scope here.

## Migration

- **`position: all`** → wrap in `all: [...]` with one leaf per slot. Mechanically codemod-able once #235 lands and `playerCount` is known at the call site.
- **`position: any`** → wrap in `any: [...]` with one leaf per slot. Codemod-able similarly.
- **`position: percentAgreement`** → no automatic migration. The semantics (numeric agreement rate against a threshold) don't reduce to a boolean tree. Authors with existing percentAgreement conditions must either:
  - rewrite the condition with explicit per-player checks against a known target value (loses the "rate against arbitrary threshold" semantics), or
  - drop the condition until a future countables / aggregates operator family lands (see below).

The viewer's preflight error path should surface clear messages for files still using the dropped values, with the migration recipe inline.

## Out of scope / future work

- **Countables and aggregates.** Operators like "at least N of these are true" (`atLeast: { n: 3, of: [...] }`), "exactly N of N", or "X% agree" are real needs, but they don't fit the boolean-tree shape and they don't fit a read-selector field. They want their own syntax, designed deliberately as a family rather than retrofitted into `position`. Tracked separately; not part of this change. `percentAgreement`'s use cases will land here when that work happens.
- **`positionSelectorSchema`** (the other `position` field, on `display` and `trackedLink`). Same field name, different role. Cleanup or rename is its own discussion.
- **Tree-level fan-out operators** like `everyPlayer:` / `anyPlayer:`. Considered and deferred — for `playerCount: 2`–`4` (the common range) the explicit tree fan-out is fine; if researchers hit verbosity at larger group sizes, these can be added non-breakingly later.

## Backward compatibility

Hard breaking change for any treatment file using `position: all`, `position: any`, or `position: percentAgreement` on a condition. Documentation and the viewer's preflight error messages need to call out the migration recipe explicitly. A codemod can handle the `all` / `any` cases mechanically; `percentAgreement` requires author judgment.

## Acceptance

- [ ] `baseConditionSchema.position` enum narrowed to `["shared", "player"]` (plus slot integer).
- [ ] `resolvedConditionSchema.position` matches.
- [ ] `PERCENT_AGREEMENT_NUMERIC_COMPARATORS` set and its associated rule removed from `validateConditionRules`.
- [ ] `validateConditionRules` game-stage `forbidSelfPosition` rule preserved and still rejects default / explicit `player` at game-stage level.
- [ ] Existing tests covering `position: all` / `any` / `percentAgreement` rewritten or removed; new tests assert the narrowed enum rejects the dropped values with helpful errors that point to the boolean-tree migration recipe.
- [ ] Annotated walkthrough updated if it uses any of the dropped values.
- [ ] CHANGELOG / release notes call out the breaking change with migration guidance.

## Related

- #235 — boolean-tree operators (`all` / `any` / `none`). Strict prerequisite for the `position: all` / `position: any` migration recipe.
- #183 — stage-level conditions (introduced the desync constraint preserved here).
- Parent syntax review (item #2 of the wider DSL audit).

JamesPHoughton:
## Summary

After #235 (boolean-tree operators), the `position` field on a condition leaf was doing too many jobs. This narrows it to a pure read selector: `"shared"`, `"player"` (default), or a slot-index integer. Drops three values:

- `position: "all"` — superseded by `all: [{position: 0}, ...]` operator nodes from #235.
- `position: "any"` — superseded by `any: [...]` operator nodes.
- `position: "percentAgreement"` — pulled out entirely. No auto-migration; a future countables/aggregates family will replace it.

The game-stage `forbidSelfPosition` rule is preserved with an updated error message that points at the boolean-tree migration recipe.

## Test plan

- [x] All vitest tests pass (728/728 in stagebook).
- [x] Lint clean.
- [x] Backward compat: `position: "shared"`, `position: "player"`, and `position: <integer>` continue to validate and evaluate identically.
- [x] Schema rejection tests for the three dropped values, with helpful error messages.
- [x] Annotated walkthrough updated; researcher docs (`conditions.md`, `syntax-reference.md`) updated; engineer docs (`architecture.md`, `platform-requirements.md`) updated.
- [ ] CI green on the PR.
- [ ] Copilot first-pass review (acting as backstop; local convergence already ran via review-subagent pass).


JamesPHoughton:
Addressed all six Copilot comments in 117335d. The substantive issue: `display.position: "all"/"any"` and similar uses on trackedLink/qualtrics urlParams still flow through `StagebookProvider.resolve()` → host `get()` via `positionSelectorSchema` (which #238 explicitly leaves alone — only condition leaves were narrowed). Removing the `"any" → "all"` normalization would have broken hosts that only implement the `"all"` storage scope, and the docs I'd written claiming hosts no longer need `"all"` were wrong. Restored the normalization, updated all three docs (StagebookProvider.tsx, architecture.md, platform-requirements.md) to correctly describe the four scopes stagebook may pass, and renamed the misleading "unknown scope" test.

Issue #278, opened 2026-05-01
Aggregate values: compute new state from multiple element responses


JamesPHoughton:
## Motivation

Several research workflows require combining data from multiple Stagebook elements into a single new value, then using that value in conditions or display:

- **Partisan affiliation** — a binary prompt (Democrat / Republican / Independent) followed by a strength prompt (Strong / Weak), combined into a single 7-point scale (-3 strong-D to +3 strong-R).
- **Running score in a game-like study** — count of `conditions` blocks that evaluated to true, used as a "points so far" metric for display or branching.
- **First-of / coalesce** — when a participant is routed through one of two prompts based on a prior answer, treat the answered one as "their response" without callers having to know which branch was taken.

Today there's no way to express "compute X from Y and Z" in a treatment file. Authors have to encode the aggregation as long disjunctive `conditions` trees, which is verbose, error-prone, and not reusable.

### Mechanism

A new non-rendering element type (working name **`aggregate`** — see open questions for the prefix word). Lives in `elements:` like every other element, declares its inputs as references, recomputes and `save()`s its value whenever an input changes.

**Element-type framing (vs. a top-level `aggregates:` side channel).** Treating aggregates as element types — even though they don't render — keeps them inside the existing scoping primitives. `conditions:` on an aggregate gates *when* it computes (e.g. "wait until all inputs have been saved"); `displayTime` / `hideTime` gate computation to a window; placement in a particular stage scopes its lifecycle to that stage's mount. These aren't accidental side effects; they're the same primitives an author would otherwise reinvent for a side-channel design.

Position-related fields (`position`, `showToPositions`, `hideFromPositions`) inherit from `elementBaseSchema` and remain syntactically allowed, but have no effect for non-rendering elements. No warning needed — gating *which players compute* is still expressible via the same mechanism, since a non-rendering element gated to position 0 simply means only player 0 runs the effect.

```yaml
- type: aggregate
  name: partisanScore
  operator: lookup
  inputs:
    - prompt.q_party.value
    - prompt.q_strength.value
  table:
    - inputs: [democrat, strong]
      value: -3
    - inputs: [democrat, weak]
      value: -2
    # ...etc
```

Renders nothing visible. References to it work like any other source: `aggregate.partisanScore.value` resolves through the same dotted-path resolver, hitting storage key `aggregate_partisanScore`.

### Eager computation; always player-scoped

[...]

Issue #277, opened 2026-05-01
Cross-file template imports for reusable measurement instruments


JamesPHoughton:
## Motivation

Researchers often develop Stagebook syntax for a measurement instrument — a survey, a battery of prompts, a scoring rule — that they want to use across multiple studies. Today there's no way to share these: the entire `templates:` block has to be copy-pasted between treatment files, and any update to the instrument has to be replicated by hand.

We want a way to ship a measurement instrument as a self-contained folder (a treatment file with `templates:` only, plus its required prompt files), drop the folder into a study, and import its templates by reference.

### Mechanism

A new top-level `imports:` field on the treatment file:

```yaml
imports:
  - ./surveys/tipi/tipi.stagebook.yaml
  - ./scoring/partisan-7pt/scoring.stagebook.yaml

templates:
  # ...inline templates...
```

The path is relative to the importing treatment file.

### Pipeline (three new Stagebook helpers; host owns the loading loop)

Imports are resolved + path-rewritten *before* any other Stagebook step. The schema sees fully expanded templates and validates them as today. The new pipeline introduces three helpers in the React-free entry point and one new host responsibility (a loading loop):

```typescript
import {
  parseTreatmentYaml,
  resolveImportPath,
  resolveImports,
  fillTemplates,
  treatmentFileSchema,
} from "stagebook";

// 1. Parse the main file. Returns { parsed, imports } — the import list
//    surfaces here so the host never has to peek into Stagebook's schema
//    to find import declarations.
const { parsed: main, imports: mainImports } =
  parseTreatmentYaml(mainYamlString);

// 2. Host loop: load every (transitively) imported file. The host owns
//    sync vs async, error handling, and any extra canonicalization
//    (symlinks, case folding) it wants. Stagebook owns syntactic path
//    normalization via resolveImportPath.
const loaded = new Map<string, ParsedTreatmentFile>();
const queue = mainImports.map((p) =>
  resolveImportPath("./main.yaml", p),
);

while (queue.length) {
  const path = queue.shift()!;
  if (loaded.has(path)) continue;                   // dedup
  const text = await loadFile(path);                // host's loader
  const { parsed, imports } = parseTreatmentYaml(text);
  loaded.set(path, parsed);
  queue.push(...imports.map((p) => resolveImportPath(path, p)));
}

// 3. Stagebook merges + path-rewrites + cycle-detects across the full set.
const templates = resolveImports({ main, files: loaded });

[...]

PR #283, opened 2026-05-06
Disallow storage-key reuse across all phases of a treatment (#281)

### LINKED ISSUE #281: Disallow storage-key reuse across all phases of a treatment
## Motivation

The storage-key collision detector from #149 catches collisions within a single game stage, within a single intro sequence, and within a single exit sequence — but **not across game stages, and not across phases**. The carve-out was deliberate: cross-stage reuse of a prompt name was treated as a legitimate "tracking change over time" pattern (pre-test in stage 1, post-test in stage 5).

In light of #277 (cross-file template imports) and #278 (aggregate values), that carve-out is the wrong default. Module reuse — where a researcher imports the same instrument multiple times — makes accidental cross-stage collision much more likely than deliberate longitudinal reuse. A researcher who imports TIPI in stage 1 and forgets to change the prefix when importing it again in stage 5 gets silent overwrites with no warning.

The convenience case (longitudinal reuse of the same name) is not strictly needed — the existing per-element `name:` override lets authors disambiguate explicitly:

```yaml
# Stage 1 (pre-test)
- type: prompt
  file: tipi/q1.prompt.md
  name: pretest_tipi_q1

# Stage 5 (post-test)
- type: prompt
  file: tipi/q1.prompt.md
  name: posttest_tipi_q1
```

Slight verbosity, large safety win.

## Proposal

The storage-key namespace is **global within a treatment file**. The detector emits a single collision list spanning every game stage, every intro step, every exit step, every aggregate element, and every other element type that produces a `{type}_{name}` key.

Open question: error vs warning. PR #149 emits warnings to allow graceful degradation. Disallowing reuse argues for promoting these to errors at validation time — the failure is unambiguous, the fix is simple, and the silent-overwrite footgun is closed. Recommend **error**, but capture this as a decision needed.

## Implementation

Extend `collectStorageKeyWarnings` (or its successor) to walk all phases in one pass instead of per-phase. The detection logic (matching `{type}_{name}` keys, skipping unnamed elements with non-deterministic fallbacks) doesn't change — only the scope.

Authors who need the same prompt file in multiple places use the existing `name:` override.

## Decisions needed before implementation

- [ ] **Error vs warning.** Recommend error.
- [ ] **Backward compatibility.** Existing treatment files that legitimately reuse names across stages will start failing. Worth a deprecation period (warning first, error after one minor version), or a clean break? Recommend clean break — the legitimate cases are easy to fix and the safety win is immediate.
- [ ] **Whether the rule applies to all element types or only those known to produce overwriting keys.** Aggregate elements (#278) overwrite by design; prompt/survey/timeline are append-mode-friendly in some hosts. Recommend uniform rule — simpler to reason about, no per-type carve-outs.

## Out of scope

- Changing the `{type}_{name}` storage-key derivation itself.
- Adding any cross-treatment collision detection (treatments are independent units).

## Related

- #149 — original storage-key collision detector. This issue widens its scope.
- #277 — cross-file template imports. The convention-based prefix extension relies on collision detection as the safety net; this issue makes that safety net work for the most natural module-reuse case.
- #278 — aggregate values. Aggregates always overwrite, so cross-stage reuse of an aggregate name is unambiguously wrong.

JamesPHoughton:
Closes #281. Supersedes #149.

## Summary

- New `collectStorageKeyCollisions(data)` helper at [packages/stagebook/src/schemas/storageKeyCollisions.ts](packages/stagebook/src/schemas/storageKeyCollisions.ts) builds a single global key map across every phase of a treatment file (intro sequences, all game stages, exit sequence) and returns duplicates. No per-stage carve-outs.
- Wired into `treatmentFileSchema`'s `superRefine` so every host that validates against the schema gets the check uniformly. Collisions surface as Zod errors, not warnings.
- Authors who need the same prompt file in multiple places use the per-element `name:` override (`name: pretest_q1` vs `name: posttest_q1`) to disambiguate.

## Why this changes

The previous design (PR #149) treated cross-stage reuse as a legitimate "tracking change over time" pattern and emitted warnings only within a single stage / intro sequence / exit sequence. With #277 (template imports) and #278 (aggregate values) on the way, that carve-out becomes a silent-overwrite footgun for the most natural module-reuse case (a researcher importing TIPI in stage 1 and again in stage 5 with the same prefix). The longitudinal pattern is recoverable via explicit `name:` overrides; the safety win for module reuse and aggregates is much larger.

## Files

- `storageKeyCollisions.ts` — new module (treatment-wide collection logic, ports key-derivation from #149)
- `storageKeyCollisions.test.ts` — 13 new tests covering within-stage, cross-stage, cross-phase, multi-element-type, and false-positive cases
- `treatment.ts` — superRefine integration + import
- `treatment.test.ts` — fixture rename: the "validate entire file" fixture happened to reuse `namedPrompt` across an intro step and a game stage; renamed the intro prompt to `introNamedPrompt` (exactly the migration real authors will apply)
- `schemas/index.ts` — re-export the helper for hosts that want direct access

## Migration

Treatment files with cross-stage name reuse will start failing validation. The fix is mechanical: rename the duplicates with phase-specific `name:` overrides. The error message points at exactly which key collides and how many elements share it.

## Test plan

- [x] `npm test` — stagebook 849 / viewer 85 / vscode 189 — all pass.
- [x] `npm run lint` — clean.
- [x] Cross-stage / cross-phase collision detection covered by 13 unit tests.
- [x] Existing vscode `unique|duplicate` warning classifier (in `validateTreatment.ts`) is unaffected — only YAML-extraction errors go through that path; schema errors come through Zod and are correctly classified as errors.

### DIFF
--- a/packages/stagebook/src/schemas/index.ts
+++ b/packages/stagebook/src/schemas/index.ts
@@ -100,6 +100,11 @@ export {
 } from "./safeParseTreatmentFile.js";
 
 export {
+  collectStorageKeyCollisions,
+  type StorageKeyCollision,
+} from "./storageKeyCollisions.js";
+
+export {
   resolvedElementSchema,
   resolvedStageSchema,
   resolvedIntroExitStepSchema,
--- a/packages/stagebook/src/schemas/storageKeyCollisions.test.ts
+++ b/packages/stagebook/src/schemas/storageKeyCollisions.test.ts
@@ -281,6 +281,83 @@ describe("collectStorageKeyCollisions", () => {
     expect(collisions[0].key).toBe("audio_intro.mp3");
   });
 
+  it("derives keys for mediaPlayer (name OR file fallback) — matches Element.tsx runtime", () => {
+    // Per Element.tsx:277, mediaPlayer falls back to the raw `file` field
+    // (NOT `url` — that field was renamed to `file` in #249).
+    const data = {
+      treatments: [
+        {
+          name: "t1",
+          gameStages: [
+            {
+              name: "stage1",
+              elements: [
+                { type: "mediaPlayer", file: "intro.mp4" },
+                { type: "mediaPlayer", file: "intro.mp4" },
+              ],
+            },
+          ],
+        },
+      ],
+    };
+    const collisions = collectStorageKeyCollisions(data);
+    expect(collisions).toHaveLength(1);
+    expect(collisions[0].key).toBe("mediaPlayer_intro.mp4");
+  });
+
+  it("flags any two qualtrics elements in the same scope as colliding (fixed key)", () => {
+    // Qualtrics writes to a fixed `qualtricsDataReady` key regardless of
+    // name/url (Qualtrics.tsx:50), so any two qualtrics elements in the
+    // same scope silently overwrite each other.
+    const data = {
+      treatments: [
+        {
+          name: "t1",
+          gameStages: [
+            {
+              name: "stage1",
+              elements: [
+                { type: "qualtrics", url: "https://example.qualtrics.com/a" },
+                { type: "qualtrics", url: "https://example.qualtrics.com/b" },
+              ],
+            },
+          ],
+        },
+      ],
+    };
+    const collisions = collectStorageKeyCollisions(data);
+    expect(collisions).toHaveLength(1);
+    expect(collisions[0].paths).toHaveLength(2);
+  });
+
+  it("flags qualtrics elements as colliding even across stages within a treatment", () => {
+    // Different URLs, different stages — still collides because the key is fixed.
+    const data = {
+      treatments: [
+        {
+          name: "t1",
+          gameStages: [
+            {
+              name: "stage1",
+              elements: [
+                { type: "qualtrics", url: "https://example.qualtrics.com/a" },
+              ],
+            },
+            {
+              name: "stage2",
+              elements: [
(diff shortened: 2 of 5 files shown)


JamesPHoughton:
The Copilot review thread stalled before posting, but the user shared the in-progress notes. Two real bugs surfaced — both pushed in `2e749fc`:

1. **`mediaPlayer` key derivation used `url` instead of `file`.** The schema field was renamed `url` → `file` in #249, and [Element.tsx:277](https://github.com/deliberation-lab/stagebook/blob/main/packages/stagebook/src/components/Element.tsx#L277) falls back to the raw `file` field. Two unnamed mediaPlayer elements with the same file weren't flagged before this fix.

2. **`qualtrics` wasn't handled at all.** [Qualtrics.tsx:50](https://github.com/deliberation-lab/stagebook/blob/main/packages/stagebook/src/components/elements/Qualtrics.tsx#L50) writes to a fixed `qualtricsDataReady` key regardless of name/url — so any two qualtrics elements in the same scope silently overwrite each other. Encoded as a synthetic constant key so the existing duplicate logic catches them.

Three new tests: mediaPlayer file-fallback, two-qualtrics-same-stage, two-qualtrics-different-stages-same-treatment. 18 collision tests total.

Other review notes were either stylistic (path precision matches existing patterns), already-settled-by-design (cross-treatment scoping, pre-fill vs post-fill validation), or speculative (Copilot rambling about hypothetical scenarios).

PR #285, opened 2026-05-06
Numeric values for multipleChoice prompts (#282)

### LINKED ISSUE #282: Numeric values for radio prompts
## Motivation

Sliders today let authors specify per-option numeric values via `- <number>: <label>` syntax, which downstream conditions and analyses can compute over directly. Radio (multiple-choice) prompts only support text labels; their stored `value` is always a string. To do numeric aggregation over Likert-style radio prompts (story 5 in #278's user-stories list — composite scale scoring), authors today have to write a `case` cascade mapping `"Strongly disagree" → 1`, `"Disagree" → 2`, etc., for every prompt. Verbose and error-prone.

Reusing the slider syntax for radio prompts removes that mapping step. `average` / `sum` aggregates over the prompts work directly. Most validated psychometric instruments (TIPI, RWA, SDO, BFI, Need-for-Cognition, etc.) use this pattern.

## Scope

Radio and slider prompts only. Checkbox is awkward (averaging over multi-select doesn't have clean semantics); open-response and free-text aren't candidates.

### Syntax

Radio prompts gain the existing slider syntax. Per option:

```
- <number>: <label>     # numeric value with display label
- <label>               # text-only (existing behavior, unchanged)
```

### Strict per-prompt mode

Every option in a single prompt must be either all-numeric or all-text. Mixed = validation error. Catches authoring mistakes early; avoids silent heterogeneous-type responses.

### Storage shape

Saved response objects gain a `label` field, complementing the existing `value`:

| Prompt | Source line | `value` | `label` |
|---|---|---|---|
| Radio (numeric) | `- 1: Strongly disagree` | `1` (number) | `"Strongly disagree"` |
| Radio (numeric, bare number) | `- 1` | `1` (number) | `"1"` (stringified) |
| Radio (text-only) | `- Strongly disagree` | `"Strongly disagree"` | `"Strongly disagree"` |
| Slider | `- 1: Strongly disagree` or `- 1` | `<chosen number>` (number) | not saved |

**`label` is always a string for radio responses** (never null, never numeric). In text-only mode it duplicates `value`; in numeric mode with a label it's the author-supplied text; in numeric mode with a bare number it's the stringified number. Consumers that care about labels read `label` regardless of mode.

**Sliders save `value` only.** Sliders are continuous-position inputs — a participant can land at 4.3 between labeled ticks; the labels in the prompt definition are display annotations on certain positions, not the response itself. Saving a label for a slider response would misrepresent what was measured. The label-to-position mapping stays in the prompt definition for the slider UI; hosts that want to display "you said 'Strongly disagree'" when a participant happens to land exactly on a labeled tick can do that lookup against the prompt definition at display time.

**Both radio and slider accept bare numbers**, but they mean different things. A slider with `- 5` is a position-only tick with no annotation (the slider UI shows position 5 with no text below). A radio with `- 5` is a button labeled "5" (the participant sees and clicks the option whose visible label is "5"). Slider responses don't save a label; radio responses save the stringified number as the label, so `label` is always a string for radio.

A researcher using a numeric-only scale (e.g., 1–7 satisfaction with no text descriptions) writes:

```
- 1
- 2
- 3
- 4
- 5
- 6
- 7
```

Renders as 7 numbered radio buttons; saves `value: <chosen number>, label: "<chosen number>"`.

### Backward compatibility

- Existing text-only radio prompts: no change to `value`. New `label` field equals `value`. Existing references to `prompt.X.value` keep working.
- Existing slider prompts: no change. Slider responses continue to save `value` only.
- A prompt that an author deliberately migrates from text to numeric mode: `value` changes from string to number. References that expected a string break. This is intentional — the mode change is an authoring decision, not a silent conversion.

### Validation

Per-prompt rules:
- All-numeric or all-text (no mixed). Mixed → validation error. Mode is inferred per prompt: any option matching `<number>` or `<number>: <label>` → numeric mode; otherwise text mode.
- Numeric values must be unique within the prompt (no duplicate codes for different options).
- Numeric values do not need to be ordered or evenly spaced (e.g., `-1: Democrat / 1: Republican / 0: Independent` is valid).
- Bare numbers are accepted in both radio and slider. In radio, the visible button label is the stringified number; in slider, the tick is unannotated.

Sliders already enforce numeric values are unique and orderable; that logic carries over to radio with the ordering check relaxed.

## Connection to #278 (aggregate values)

With this, the most common aggregate use cases (Likert composite scoring, partisan-scale-style branching) collapse from "case-cascade-mapping-text-to-number, then average" into just "average over the prompts directly." The aggregate vocabulary doesn't change; the *need* to use `case` for routine workflows decreases significantly.

This is not a prerequisite for #278 — case cascades can still do the mapping today — but it materially improves the ergonomics of the most common aggregate patterns. Recommend implementing alongside or before #278's first real use cases.

## Decisions needed before implementation

- [ ] **Field name for the display text.** Working name: `label`, matching the existing slider parser's internal naming. Alternative: `displayText` (more explicit, longer).
- [ ] **Whether to expose `label` even in text-only radio mode** (where it's a duplicate of `value`). Recommend yes — gives consumers a single field name to read regardless of radio mode.
- [ ] **Whether the numeric values must be integers, or any finite number.** Sliders allow any finite number today; radio Likert is typically integer-coded but no strong reason to restrict. Recommend allow any finite number.
- [ ] **Whether to add an explicit `numeric: true | false` field on the prompt metadata.** Today the mode is inferred from option contents (any line with `<number>: <label>` → numeric). Inferring is simpler; explicit declaration catches mode mismatches earlier. Recommend infer for v1.

## Out of scope for v1

- Checkbox prompts. Multi-select numeric aggregation has unclear semantics (sum? average? max?); defer until a use case shows up.
- Open-response and free-text prompts.
- Per-option metadata beyond `value` + `label` (e.g., colors, icons, exclusivity flags).

## Related

- #278 — aggregate values. The most common aggregate use cases benefit directly from this change.
- #243 — prompt-file format (the change this builds on).

JamesPHoughton:
Closes #282.

## Summary

`multipleChoice` prompts gain the slider's `<number>: <label>` syntax for specifying per-option numeric values. The most common aggregate use cases (Likert composite scoring, partisan-style branching, TIPI-family scales) collapse from "case cascade mapping text → number, then average" into "average over the prompts directly" — the values are already numeric in the response.

## What's allowed

```yaml
---
type: multipleChoice
---
How much do you agree?
---
- 1: Strongly disagree
- 2: Disagree
- 3: Neutral
- 4: Agree
- 5: Strongly agree
```

Bare numbers are also accepted (`- 1` renders as a button labeled `"1"`):

```
- 1
- 2
- 3
```

## Rules

- **All-or-nothing per prompt.** Every option in a single prompt must be either all-numeric or all-text. Mixed = validation error.
- **Numeric values must be unique within the prompt.** Duplicates = validation error.
- **Single-select only in v1.** Numeric mode + `select: multiple` is rejected.
- **`label` is always a string** for multipleChoice responses.

## Storage shape (multipleChoice)

| Source line | `value` | `label` |
|---|---|---|
| `- 1: Strongly disagree` | `1` (number) | `"Strongly disagree"` |
| `- 1` | `1` (number) | `"1"` |
| `- Strongly disagree` | `"Strongly disagree"` | `"Strongly disagree"` |

Sliders save `value` only (continuous input, no honest label) — no change from previous behavior.

## Schema changes

- `ParsedPromptFile` gains `responsePoints: number[]` (canonical name for numeric per-option values).
- `sliderPoints` retained as a deprecated alias — identical to `responsePoints` for sliders, empty otherwise (matches pre-#282 behavior, no consumer break).
- `parseSliderLine` renamed to `parseNumericResponseLine`; new `isNumericResponseLine` helper drives mode detection.

## Tests

9 new tests in [promptFile.test.ts](packages/stagebook/src/schemas/promptFile.test.ts):
- numeric `<number>: <label>` parsing
- bare-number parsing (label = stringified number)
- non-integer values
- text-only mode (no regression)
- mixed numeric+text rejection
- duplicate-value rejection
- multi-select numeric rejection
- multi-select text mode still works
- slider populates both `responsePoints` and (deprecated) `sliderPoints`

## Test plan

- [x] `npm test` — stagebook 863 / viewer 85 / vscode 189 — all pass.
- [x] `npm run lint` — clean.
- [x] `npx prettier --check` — clean.
- [ ] Manual: open a numeric multipleChoice prompt in the viewer, click an option, verify the saved record has `value` as a number and `label` as the displayed text.

## Notes

This unblocks the most common patterns from #278 (aggregate values) — Likert composites and partisan branching no longer need `case` cascades to convert text labels to numbers.

### DIFF
--- a/packages/stagebook/src/components/Element.tsx
+++ b/packages/stagebook/src/components/Element.tsx
@@ -206,7 +206,7 @@ export function Element({ element, onSubmit, stageDuration }: ElementProps) {
           </p>
         );
       }
-      const { metadata, body, responseItems, sliderPoints } = parsed.data;
+      const { metadata, body, responseItems, responsePoints } = parsed.data;
       const promptName =
         element.name ?? `${progressLabel}_${metadata.name ?? element.file}`;
 
@@ -220,7 +220,7 @@ export function Element({ element, onSubmit, stageDuration }: ElementProps) {
           metadata={metadata}
           body={body}
           responseItems={responseItems}
-          sliderPoints={sliderPoints}
+          responsePoints={responsePoints}
           name={promptName}
           file={element.file}
           shared={element.shared}
--- a/packages/stagebook/src/components/elements/Prompt.ct.tsx
+++ b/packages/stagebook/src/components/elements/Prompt.ct.tsx
@@ -513,3 +513,78 @@ test.describe("Multiple Choice option order", () => {
     expect(order).toEqual([...customOrderOptions]);
   });
 });
+
+// ---------------------------------------------------------------------
+// #282 — shuffle + numeric mode must keep label/point pairing intact
+// ---------------------------------------------------------------------
+//
+// Bug surfaced during review: a previous implementation shuffled the
+// labels but kept `numericPoints` in source order, so a shuffled radio
+// would record the wrong number for the chosen label. This test pins
+// the invariant that whichever label the participant picks, the saved
+// numeric value is the one originally aligned with it in the source.
+
+const NUMERIC_LIKERT_LABELS = [
+  "Strongly disagree",
+  "Disagree",
+  "Neutral",
+  "Agree",
+  "Strongly agree",
+] as const;
+const NUMERIC_LIKERT_POINTS = [1, 2, 3, 4, 5] as const;
+
+const numericShuffledMultipleChoice = {
+  metadata: {
+    name: "projects/example/numericShuffled.md",
+    type: "multipleChoice" as const,
+    shuffle: true,
+  },
+  body: "# How much do you agree?",
+  responseItems: [...NUMERIC_LIKERT_LABELS],
+  responsePoints: [...NUMERIC_LIKERT_POINTS],
+};
+
+test.describe("Multiple Choice numeric mode + shuffle (#282)", () => {
+  test("shuffled labels stay paired with their original numeric points", async ({
+    mount,
+  }) => {
+    const saved: { key: string; value: unknown }[] = [];
+    const component = await mount(
+      <Prompt
+        {...numericShuffledMultipleChoice}
+        name="testNumericShuffle"
+        value={undefined}
+        progressLabel="game_0_numeric_shuffle"
+        save={(key, value) => {
+          saved.push({ key, value });
+        }}
+        getElapsedTime={() => 0}
+      />,
+    );
+
+    // Find the radio whose displayed label is "Strongly agree" (originally
+    // paired with point 5) and click it. Whatever its display position
+    // after shuffle, the saved numeric value must be 5.
+    const radios = component.locator('input[type="radio"]');
+    const labels = await component
+      .locator("label")
+      .evaluateAll((nodes) => nodes.map((n) => (n.textContent ?? "").trim()));
+    const stronglyAgreeIdx = labels.findIndex((t) =>
+      t.includes("Strongly agree"),
+    );
+    expect(stronglyAgreeIdx).toBeGreaterThanOrEqual(0);
+    // `.click()` not `.check()` — RadioGroup is controlled, so the input's
+    // `checked` attribute won't update without a parent re-render. We only
+    // care that onChange fires with the right value.
+    await radios.nth(stronglyAgreeIdx).click();
+
+    // Wait for the 50ms debounce to fire.
+    await component.page().waitForTimeout(80);
+
+    expect(saved.length).toBeGreaterThan(0);
+    const last = saved[saved.length - 1];
+    const record = last.value as { value: unknown; label: unknown };
+    expect(record.value).toBe(5);
+    expect(record.label).toBe("Strongly agree");
+  });
+});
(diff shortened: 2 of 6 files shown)


JamesPHoughton:
The Copilot review session stalled before posting (same as #283). The session URL is private to GitHub's web UI, but I had a subagent independently audit the diff for the same kind of concrete bugs. It surfaced one merge-blocker:

**Shuffle + numeric mode misaligned labels and points.** The prior implementation shuffled `responses` (labels) at [Prompt.tsx:88-99](https://github.com/deliberation-lab/stagebook/blob/feat/issue-282-multipleChoice-numeric/packages/stagebook/src/components/elements/Prompt.tsx#L88-L99) but kept `numericPoints` in source order. A shuffled numeric Likert would record the wrong number for the chosen label (participant clicks "Strongly agree" → saved as `value: 1` instead of `5`).

Fixed in `ac94845` by tracking a single `shuffleOrder: number[]` and deriving both labels and points from it — the pairing is now structurally guaranteed at every display position. New CT test in `Prompt.ct.tsx` pins the invariant.

Other items the subagent flagged were either intentional cleanup (dropping `> ` blockquote handling for multipleChoice — already rejected by the per-type marker check at `promptFile.ts:343-360`) or pre-existing behavior unaffected by this PR.

PR #303, opened 2026-05-08
Catch Space anywhere on the page when a player is mounted (#300)

### LINKED ISSUE #300: Spacebar pauses playback but also scrolls the page
## Summary

On a page with a MediaPlayer + Timeline, pressing the spacebar to pause playback also scrolls the page.

## Steps to reproduce

1. Open a stage that contains a MediaPlayer and at least one Timeline, with enough content below the player that the page is scrollable.
2. Start playback (e.g. click the play button).
3. Press the spacebar to pause.

**Expected:** playback pauses, page does not move.
**Actual:** playback pauses *and* the page scrolls down by roughly one viewport (the browser's default Space-scroll behavior fires alongside the pause).

## Hypothesis

Both keyboard handlers we own already `preventDefault()` on Space:

- Timeline: [keyboardActions.ts:83](packages/stagebook/src/components/elements/timeline/keyboardActions.ts#L83) maps Space → `togglePlayPause`, and [Timeline.tsx:567](packages/stagebook/src/components/elements/Timeline.tsx#L567) calls `e.preventDefault()` + `e.stopPropagation()` for any returned action.
- MediaPlayer: [MediaPlayer.tsx:725-731](packages/stagebook/src/components/elements/MediaPlayer.tsx#L725-L731) handles Space and calls `preventDefault()`.

So the scroll most likely comes from a focus state where *neither* handler runs but something else still pauses the video. Suspects to check:

- Focus is on the play/pause `<button>` inside [controls.tsx](packages/stagebook/src/components/elements/mediaPlayer/controls.tsx) (after the user clicked it to start). Space activates the button (pause), but the keydown reaches `document` first and scrolls before the button activates on keyup — confirm whether the button's default Space handling is preventing scroll or not.
- Focus has moved off the player/timeline entirely (e.g. user clicked elsewhere), but the user perceives Space as still pausing because some other element re-routes it. Need to check if there's any document-level Space handler.
- The Timeline's `onKeyDown` is on a div with `tabIndex` — if focus is on a sibling/parent, the Space event may not bubble through it.

## Why this matters

Participants pause to read instructions or think; jarring scroll-jumps disrupt their focus and can scroll engagement-tracked content past the viewport, polluting measurements.

## Repro environment

- Stagebook 0.9.6 (current `main`)
- Reported in the VS Code preview, but the underlying handlers are shared with all hosts.

JamesPHoughton:
Closes #300.

## Summary
- `Timeline` and `MediaPlayer` already `preventDefault` on Space when focused, but the bug shows up when focus has drifted off the player (typically: user clicks "play", controls hide-on-play unmounts the focused button, focus falls back to body — now Space scrolls the page instead of pausing).
- `PlaybackProvider` now mounts a window-level `keydown` handler that toggles the most recently registered playback handle on bare Space, and `preventDefault`s the browser's scroll. It defers via `e.defaultPrevented` so existing focused handlers keep first dibs (no double-toggle), and skips text inputs / `[contenteditable]` / modifier-key combos so typing and OS shortcuts pass through.

## Test plan
- [x] `npx vitest run packages/stagebook/src/components/playback/` (11/11, including 9 new for #300)
- [x] `npx eslint` on changed files (clean)
- [ ] Visually verify in VS Code preview: stage with MediaPlayer + Timeline, click around to lose focus, then press Space — playback toggles, page does not scroll

### DIFF
--- a/packages/stagebook/src/components/elements/MediaPlayer.tsx
+++ b/packages/stagebook/src/components/elements/MediaPlayer.tsx
@@ -9,7 +9,11 @@ import { isYouTubeURL } from "./mediaPlayer/isYouTubeURL.js";
 import { parseVTT, type CaptionCue } from "./mediaPlayer/parseVTT.js";
 import { YouTubePlayer } from "./mediaPlayer/YouTubePlayer.js";
 import { HTML5Controls, YouTubeControls } from "./mediaPlayer/controls.js";
-import { useRegisterPlayback } from "../playback/PlaybackProvider.js";
+import {
+  useRegisterPlayback,
+  useMarkActive,
+  useIsActiveSpaceTarget,
+} from "../playback/PlaybackProvider.js";
 import type { PlaybackHandle } from "../playback/PlaybackHandle.js";
 import { computeWatchedRanges } from "../../utils/watchedRanges.js";
 import {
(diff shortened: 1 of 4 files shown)


JamesPHoughton:
Closing as superseded by [#408](https://github.com/deliberation-lab/stagebook/pull/408).

#303 proposed a window-level Space handler that would forward to the MediaPlayer regardless of where focus is. We chose the more targeted fix in #408 instead: when the controls overlay auto-hides and would orphan focus to `<body>`, focus is transferred to the MediaPlayer container (`tabIndex=0`) so the existing per-component keydown handler stays live.

This avoids the assumptions a global handler would make (single player on the page, no other Space-consumers), and keeps focus a visible affordance now that MediaPlayer has an explicit `:focus-within` ring (#405).

The current behavior was confirmed in manual testing — closing as done.

PR #316, opened 2026-05-10
Position-based fallback for unnamed audio/mediaPlayer/survey + post-fill error surfacing

JamesPHoughton:
## Summary

Hits two related authoring-UX problems surfaced by a real-world file (`dialogue-levers/pilot_3.stagebook.yaml`):

1. The file has 220+ treatments, most playing the same unnamed `counter_bell.mp3` audio in two stages. The runtime gave both occurrences the same storage key (`audio_<file>` fallback), and the validator flagged 440 collision errors. The researcher didn't actually want save data for the bell — they just wanted the sound to play.

2. The preview command silently swallowed those 440 errors and showed a generic "Could not parse treatment file for preview." popup, leaving no actionable path forward.

### Position-based runtime key for unnamed elements

`Element.tsx` now generates the runtime name as `${progressLabel}_${file}` (or `${progressLabel}_${surveyName}` for surveys) when `name:` is omitted. So two unnamed plays of the same sound in different stages get distinct storage keys naturally — no collision, both saved. Researchers opt in to cross-stage tracking by setting `name:` explicitly.

Affects audio, mediaPlayer, survey. Named elements unchanged.

### Storage-key collision detector relaxed

`storageKeyCollisions.ts` no longer flags collisions on the file-fallback key. For audio/mediaPlayer/survey, the detector returns null when `name:` is absent — researchers who don't name are opting out of cross-stage tracking; the runtime guarantees uniqueness via the position fallback. Two `name: bell` audios still collide as expected.

### Preview popup surfaces real errors

`parseTreatmentForPreview` now returns `{ ok: false, error }` with a specific message instead of `null`:
- YAML parse failure: includes the parser's error
- Import file not found: includes the canonical path
- Template expansion failure: includes the engine's error
- Schema validation failure: shows the first 3 issues with paths, plus a count of remaining ones

Both initial-preview and refresh paths show the result via `showErrorMessage(..., { modal: true })`.

### In-editor banner for post-fill validation issues

`validateTreatmentFile` now adds a single line-1 banner diagnostic when the file's expanded form has issues:

> "Stagebook: 440 validation issues in this file's expanded form. Run \"Stagebook: Preview Expanded Templates\" to see them at their expanded-line positions."

The expand-templates virtual document already runs full post-fill validation and attaches diagnostics to the virtual URI with positions naturally mapped to the expanded text. The banner just tells the researcher where to look — sidesteps the sourcemap-back-to-source problem entirely (templates multiply structure; mapping post-fill paths to source positions is hard).

If template expansion itself fails, the banner surfaces the error message directly with severity Error.

## Test plan

- [x] `pilot_3.stagebook.yaml` now parses cleanly (was 440 errors → 0)
- [x] 1221 tests pass across all workspaces (945 stagebook + 87 viewer + 189 vscode)
- [x] Updated collision detector tests cover both the named-still-collides and unnamed-no-longer-collides cases
- [x] Builds clean
- [x] VS Code extension reinstalled locally; preview popup now shows specific errors with line/path

## Out of scope (future work)

- Source-position squiggles for individual post-fill issues (sourcemapping post-fill paths back to source). Banner + expand-preview gives users a clear path to investigate; precise source squiggles would need fillTemplates instrumentation to track provenance.

### DIFF
--- a/apps/vscode/src/extension.ts
+++ b/apps/vscode/src/extension.ts
@@ -487,9 +487,11 @@ class ExpandedTemplatesProvider implements vscode.TextDocumentContentProvider {
       return `# Template expansion error:\n${commentedError}`;
     }
 
-    // Attach schema-validation diagnostics to the expanded preview URI.
-    // Positions reference the full expanded YAML; diagnostics past the
-    // truncation line still appear in the Problems panel.
+    // Schedule diagnostic-attachment for after VS Code has opened the
+    // document for this URI. Setting them synchronously inside this
+    // content-provider call is unreliable: VS Code can drop the
+    // diagnostics during document creation. setTimeout(0) defers
+    // until after the open completes.
     const vscodeDiagnostics = result.diagnostics.map((d) => {
       const diag = new vscode.Diagnostic(
         toVscodeRange(d.range),
@@ -499,7 +501,9 @@ class ExpandedTemplatesProvider implements vscode.TextDocumentContentProvider {
       diag.source = "stagebook";
       return diag;
     });
-    this.diagnostics.set(uri, vscodeDiagnostics);
+    setTimeout(() => {
+      this.diagnostics.set(uri, vscodeDiagnostics);
+    }, 0);
 
     return result.yaml;
   }
--- a/apps/vscode/src/extension.ts
+++ b/apps/vscode/src/extension.ts
@@ -487,11 +487,12 @@ class ExpandedTemplatesProvider implements vscode.TextDocumentContentProvider {
       return `# Template expansion error:\n${commentedError}`;
     }
 
-    // Schedule diagnostic-attachment for after VS Code has opened the
-    // document for this URI. Setting them synchronously inside this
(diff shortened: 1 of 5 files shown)


JamesPHoughton:
Superseded by:

- #322 — preview command surfaces specific errors instead of the generic notification (replaces the post-fill banner approach in this PR)
- #323 — position-based storage-key fallback for unnamed audio/mediaPlayer/survey + collision-detector relaxation (the standalone runtime + detector portion of this PR, cherry-picked)
- #321 — async + import-aware validator with the diff orchestrator (supersedes the reference-checking commits ed2e10f / 0419cdf, and explicitly removes the import-suppression heuristic from 7578dd3)

The remaining don't-wipe-EXPANDED-scheme-diagnostics-on-close fix (932db93) is moot post-#322 in normal use; #321's rework will refactor this path anyway. If a real reproducer surfaces before #321 lands, we'll lift it as a small standalone fix.

Issue #386, opened 2026-05-19
ListSorter: reconsider the static position-number column


JamesPHoughton:
## Background

The current `ListSorter` renders a static left column of position numbers (`1.`, `2.`, `3.`, …) alongside the draggable rows. During the polish work in #374, the reviewer flagged this as worth rethinking.

## The question

Are the static numbers pulling their weight? A few angles:

- **Coupling cost.** The column has to match the row height exactly (`min-height: var(--stagebook-row-min-height)`) or the numbers drift away from their rows. We tolerate ~3px to account for the row's 1px border. Any future change to row padding/border has to be mirrored in the number column.
- **Semantics during drag.** As rows reorder, the row content changes position but the number column doesn't — `Alpha` was \"1.\" before the drag and is \"2.\" after, but a participant who is mid-drag sees the *destination* number, not the source. Is that what we want?
- **Alternatives.**
  - Drop the column entirely (rely on visual order alone — same as Notion / Linear).
  - Render the number *inside* each row so it travels with the row during drag.
  - Keep the column but reconsider whether `1.`, `2.`, `3.` is the right glyph — maybe a smaller muted index, or none on hover.

## Polish-sweep context

This is a follow-up to #374 (ListSorter UI polish). The polish PR keeps the existing column and just makes it stay aligned. The design question is deferred here so we can think about it without blocking that PR.

## Acceptance

Decide whether to keep, remove, or restructure the position-number column. If we keep it, document why; if we change it, file the implementation issue.

JamesPHoughton:
Closing as not planned.

Decision: **keep the position-number column as it is.**

Reasoning: the numbers stay put during drag (don't travel with the row), which on reflection is the right behavior — they label *slots*, not *items*. A participant dragging Alpha from slot 1 to slot 2 watches the row arrive at the "2." position, which is the destination they care about. The visual "the number 1 sits next to a row that's about to be in slot 2 mid-drag" reads naturally once you frame the numbers as slot labels rather than row labels.

The coupling cost (column height has to mirror row height) is a known, contained constraint — already paid for in #387 via the matching `--stagebook-row-min-height` transparent borders, with a regression test (`position-number labels stay aligned with their rows (no cumulative drift)` in ListSorter.ct.tsx) that catches future drift.

Not enough upside to justify the work.

Issue #385, opened 2026-05-19
ListSorter: replace @hello-pangea/dnd with a stagebook-owned drag implementation


JamesPHoughton:
## The CLAUDE.md violation

`packages/stagebook/src/components/form/ListSorter.tsx` imports `@hello-pangea/dnd` (a fork of react-beautiful-dnd). This violates CLAUDE.md's explicit rule:

> **No external UI library dependencies.** Don't import Radix, shadcn, etc. Use them as references to audit our implementations for accessibility and edge cases, but keep full ownership so upstream changes can't alter experiment behavior.

The import predates the polish sweep (it was there when #374 was filed; #383, #381, #378, #377, #376, #375 all closed the gap for other components without addressing this one). The polish PR (#TBD) added hover / focus-visible / touch-target / reduced-motion / glyph fixes ON TOP of the existing library, leaving the import in place.

## Why this matters specifically here

For research instruments, "stagebook X.Y produces the same participant experience as stagebook X.Y+1" is a science-data property, not just an engineering preference. Any change in @hello-pangea/dnd's drag heuristics, keyboard handling, drop animations, or announcement strings affects how participants interpret the task — and stagebook can't audit that surface without owning the code.

## What to replace it with

A stagebook-owned drag implementation should preserve at minimum:

- **Mouse drag** — click-and-drag a row to reorder
- **Touch drag** — same on mobile / tablet
- **Keyboard a11y** — `Tab` to focus a row, `Space` to lift, `ArrowUp` / `ArrowDown` to move, `Space` to drop, `Escape` to cancel (the @hello-pangea/dnd model is a reasonable starting point)
- **Live region announcements** — "Item Alpha picked up. Now in position 2 of 4." etc. Screen-reader users need this for non-visual drag feedback
- **Drop indicator / placeholder** — visual gap showing where the dragged row will land
- **`prefers-reduced-motion` handling** — skip drop animations for users who opt out

References:
- [Radix Sortable example](https://www.radix-ui.com/) (community recipe, no first-party primitive)
- [shadcn sortable](https://shadcn-extension.vercel.app/docs/sortable)
- [HTML Drag and Drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) — for the mouse path
- [react-aria useSortable](https://react-spectrum.adobe.com/react-aria/useSortable.html) — for the a11y patterns (reference, don't import)

## Acceptance

[...]

Issue #373, opened 2026-05-18
UI polish: Button


JamesPHoughton:
Sub-issue of #350 — apply the UI polish checklist to `Button`.

## Per-component checklist (from #350)

- [ ] Hover affordance on the full clickable area
- [ ] Focus visibility using `--stagebook-focus-ring` (and `:focus-visible` semantics — show ring only on keyboard nav, not mouse click, where appropriate)
- [ ] Disabled state contrast
- [ ] Keyboard navigation (arrow keys, space, enter, tab order)
- [ ] ARIA roles + states (`aria-checked`, `aria-disabled`, `aria-invalid`, etc.)
- [ ] Reduced-motion behavior for any animations
- [ ] Touch-target size on small viewports

## Known gaps

- Hover / active / disabled states.
- Focus ring; `:focus-visible`.
- Touch-target size.

## Constraint

Reference shadcn / Radix UI / Apple HIG / Material Design 3 for patterns, but **don't import them** — stagebook keeps full ownership of its components so baked-in behavior keeps shipping with them. (Per CLAUDE.md.)

## Out of scope

- Theme parity with any specific host (hosts override via CSS variables).
- External UI dependencies.

Issue #372, opened 2026-05-18
UI polish: Slider


JamesPHoughton:
Sub-issue of #350 — apply the UI polish checklist to `Slider`.

## Per-component checklist (from #350)

- [ ] Hover affordance on the full clickable area
- [ ] Focus visibility using `--stagebook-focus-ring` (and `:focus-visible` semantics — show ring only on keyboard nav, not mouse click, where appropriate)
- [ ] Disabled state contrast
- [ ] Keyboard navigation (arrow keys, space, enter, tab order)
- [ ] ARIA roles + states (`aria-checked`, `aria-disabled`, `aria-invalid`, etc.)
- [ ] Reduced-motion behavior for any animations
- [ ] Touch-target size on small viewports

## Known gaps

- Recent work in #326 / #337 fixed thumb shape, edge-reach, tick alignment.
- Audit keyboard nav (arrow keys, page-up/down for big steps), ARIA `role=slider`, and the unanchored-thumb pattern's a11y story.

## Constraint

- Reference shadcn / Radix UI / Apple HIG / Material Design 3 for patterns, but **don't import them** — stagebook keeps full ownership of its components so baked-in behavior keeps shipping with them. (Per CLAUDE.md.)
- The thumb is initially hidden so as not to bias participants responses through anchoring. This behavior is intentional and should be preserved.

## Out of scope

- Theme parity with any specific host (hosts override via CSS variables).
- External UI dependencies.

PR #387, opened 2026-05-19
UI polish: ListSorter (#374)

### LINKED ISSUE #374: UI polish: ListSorter
Sub-issue of #350 — apply the UI polish checklist to `ListSorter`.

## Per-component checklist (from #350)

- [ ] Hover affordance on the full clickable area
- [ ] Focus visibility using `--stagebook-focus-ring` (and `:focus-visible` semantics — show ring only on keyboard nav, not mouse click, where appropriate)
- [ ] Disabled state contrast
- [ ] Keyboard navigation (arrow keys, space, enter, tab order)
- [ ] ARIA roles + states (`aria-checked`, `aria-disabled`, `aria-invalid`, etc.)
- [ ] Reduced-motion behavior for any animations
- [ ] Touch-target size on small viewports

## Known gaps

- Drag-and-drop a11y is famously hard; audit keyboard alternatives (move-up / move-down buttons or keyboard drag).
- Hover affordance on the drag handle vs the row.
- ARIA `aria-grabbed` / `aria-dropeffect` or the more recent live-region pattern.

## Constraint

Reference shadcn / Radix UI / Apple HIG / Material Design 3 for patterns, but **don't import them** — stagebook keeps full ownership of its components so baked-in behavior keeps shipping with them. (Per CLAUDE.md.)

## Out of scope

- Theme parity with any specific host (hosts override via CSS variables).
- External UI dependencies.

JamesPHoughton:
Fixes #374.

Sixth and final tier-1 component in the form-input polish sweep. Same per-component checklist as #370 / #371 / #372 / #373: hover affordance, keyboard-only `:focus-visible` ring, touch-target sizing, reduced-motion respect, border longhands.

## Visible changes

- **Drag-handle glyph: `⇅` → `⠿`** (Braille drag handle). `⇅` read more like a "sort" icon; `⠿` is the de-facto "drag me" affordance across modern web UI (shadcn, Notion, GitHub PR reorder). Wrapped in `<span aria-hidden="true">` so screen readers read just the item text.
- **Hover affordance on static rows** — shifts to `--stagebook-hover-bg`, the same token Radio / Checkbox rows use. Skipped while dragging so the drag-state styling isn't masked.
- **Keyboard `:focus-visible` ring** — appears when the user tabs to a row before pressing Space-to-grab; tells them they've selected the lift target. Scoped to `[data-dragging="false"]` so the ring rule and drag-state rule don't fight on specificity during an active drag.
- **Stronger drag-state shadow** — `0 4px 8px rgba(0,0,0,0.12)` so a lifted row visually "picks up" from the list, instead of the previous flat treatment.
- **Touch-target sizing** — `min-height: var(--stagebook-row-min-height)` (2.25rem default), matching Radio / Checkbox rows. Position-number labels match the same min-height so they don't drift away from their rows.
- **Reduced-motion** — `@media (prefers-reduced-motion: reduce)` disables the row hover transition.
- **Border longhands** — same `border` → `borderWidth/Style/Color` fix pattern as #367, since React's inline-style diff clears longhand and loses the shorthand expansion.

## Structural / specificity pattern

State-dependent properties (`background-color`, `border-color`, `box-shadow`, `transition`) moved from inline style to a class-scoped `<style>` block via `useId()`. Inline-style specificity wins over CSS class rules, so without this move the hover / focus-visible / reduced-motion rules wouldn't apply. Same trap I hit on Slider / TextArea / Button.

## Tests

10 ListSorter CT tests pass, including 5 new polish tests:
- row darkens on hover (when not dragging)
- focus ring appears on keyboard focus via `:focus-visible`
- rows meet touch-target sizing (≥36px tall)
- `prefers-reduced-motion`: rows disable the hover transition
- position-number labels align with their rows (same height, within 3px to account for the row's 1px border)

## Follow-ups deferred

- #385 — replace `@hello-pangea/dnd` per CLAUDE.md's no-external-UI rule. Pre-existing import, not introduced here.
- #386 — rethink the static position-number column (does it pull its weight? does it confuse mid-drag?). Polish kept the column and just made it stay aligned.

## Test plan

- [x] `npm run test:ct -w stagebook -- --grep "ListSorter"` (10/10 pass)
- [ ] Visual smoke in the viewer / component-gallery example
- [ ] CI green

### DIFF
--- a/examples/component-gallery/prompts/listsorter_demo.prompt.md
+++ b/examples/component-gallery/prompts/listsorter_demo.prompt.md
@@ -10,9 +10,9 @@ Drag to reorder. Top = most often.
 
 ---
 
-- radio: RadioGroup (pick one)
-- checkbox: CheckboxGroup (pick any)
-- select: Select (dropdown)
-- textarea: TextArea (free text)
-- slider: Slider (continuous scale)
-- listsorter: ListSorter (drag-to-rank)
+- RadioGroup (pick one)
+- CheckboxGroup (pick any)
+- Select (dropdown)
+- TextArea (free text)
+- Slider (continuous scale)
+- ListSorter (drag-to-rank)
--- a/packages/stagebook/src/components/form/ListSorter.ct.tsx
+++ b/packages/stagebook/src/components/form/ListSorter.ct.tsx
@@ -130,26 +130,31 @@ test.describe("ListSorter", () => {
     expect(transition).toBe("none");
   });
 
-  test("position-number labels align with their rows (same height)", async ({
+  test("position-number labels stay aligned with their rows (no cumulative drift)", async ({
     mount,
   }) => {
-    // Position numbers live in a separate left column; without
-    // matching `min-height` on the number labels, taller draggable
-    // rows would drift away from "1.", "2.", etc.
+    // Bug observed in the component gallery: numbers were centered
+    // on row 1 but ~10px above center by row 6 — the rows are 2px
+    // taller per item than the number labels were (1px each top +
+    // bottom border that the labels didn't have), so a six-row list
+    // drifted ~10px. The fix gives the labels matching transparent
+    // borders.
+    //
+    // This test compares vertical centers on the LAST row, which is
+    // where the cumulative drift accumulates. The previous version
+    // compared the first row with a 3px tolerance and missed this
+    // bug entirely.
     const component = await mount(<MockListSorter items={testItems} />);
-    const rowBox = await component.getByTestId("draggable-0").boundingBox();
-    // The number labels are <p> elements inside the left column.
-    // Pick the first one and compare bounding-box heights.
-    const numberBox = await component
+    const lastRowBox = await component.getByTestId("draggable-3").boundingBox();
+    const lastNumberBox = await component
       .locator("p")
-      .filter({ hasText: "1." })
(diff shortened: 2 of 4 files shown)


JamesPHoughton:
Subagent review: **SHIP** (no blockers).

Highlights:
- Pattern parity with Button / Slider / TextArea / Select is clean — `useId` + safeId regex, structural inline + state in CSS, border longhands, 120ms ease-out, reduced-motion block all on-pattern.
- `:focus-visible` and `:hover` correctly scoped to `[data-dragging="false"]`. Since static/dragging are mutually exclusive at runtime, no real specificity fight.
- `⠿` (U+283F) with `aria-hidden` is the right glyph swap — no a11y regression.
- Test tolerances (3px row vs. number label) are justified.

Acknowledged minor nits (not addressing in this PR):
1. `displayIndex` uses `[...Array(n+1).keys()].slice(1)` — pre-existing.
2. Drag state replaces the focus ring with the lifted-row shadow during keyboard drag. Intentional and called out in the CSS comment — heavier shadow + darker border serve the same "I'm holding this" purpose.

Modest test-coverage gap: no assertion that the `[data-dragging="true"]` CSS actually fires. Consistent with the predecessors (driving dnd state synthetically is awkward), not a regression.

Issue #350, opened 2026-05-14
UI polish sweep: audit components against shadcn / Radix as references


JamesPHoughton:
## Idea

Walk our component families one at a time and compare against well-trodden references (shadcn, Radix UI, Apple HIG, Material Design 3) for micro-interactions and accessibility patterns we may be missing.

The trigger was noticing that our `RadioGroup` / `CheckboxGroup` don't have the subtle hover-state fill on the entire option row — a near-universal pattern that quietly teaches participants "the whole label is the click target." There are likely a handful of similar small gaps.

## Constraint (per [CLAUDE.md](CLAUDE.md))

- **Reference, don't import.** Components stay fully owned by stagebook so baked-in behavior (keystroke stats, paste detection, debounce timing, the unanchored-slider pattern) keeps shipping with them.
- Adopt patterns, not packages.

## Suggested approach when we revisit

Bounded sweeps — one component family per PR. Likely order:

1. **Form inputs** (`RadioGroup`, `CheckboxGroup`, `Select`, `TextArea`, `Slider`, `Button`, `ListSorter`).
2. **Layout primitives** (`Stage` columns, `Markdown`, `Separator`).
3. **Conditional renders** (`TimeConditionalRender`, `PositionConditionalRender`, `ConditionsConditionalRender`, `SubmissionConditionalRender`, `StageConditionGate`).

Per-component checklist to apply uniformly:

- Hover affordance on the full clickable area
- Focus visibility (outline / ring) using `--stagebook-focus-ring`
- Disabled state contrast
- Keyboard navigation (arrow keys, space, enter, tab order)
- ARIA roles + states (`aria-checked`, `aria-disabled`, `aria-invalid`, etc.)
- Reduced-motion behavior for any animations
- Touch-target size on small viewports

## Out of scope

- Restyling for theme parity with any specific host. Hosts override via CSS variables already.
- Adding any external UI dependency. Per CLAUDE.md.

## Discovered while

Discussing whether a hover-state highlight on radio options would be worth adding — yes, and the broader observation is that we likely have a few more of these we'd see if we sat down and looked.

JamesPHoughton:
Breaking this into sub-issues per component, starting with form inputs (tier 1).

### Tier 1 — Form inputs

- [ ] #368 — UI polish: RadioGroup
- [ ] #369 — UI polish: CheckboxGroup
- [ ] #370 — UI polish: Select
- [ ] #371 — UI polish: TextArea
[...]

Issue #33, opened 2026-04-08
Markdown component: ship default heading/list styling so prompts render with visual hierarchy on hosts that reset CSS


JamesPHoughton:
## Problem

The `Markdown` component renders prompt content via `react-markdown`, which produces semantically correct HTML (`<h1>`, `<h2>`, `<p>`, `<ul>`, etc.). But it doesn't ship any default styling for those elements. On a vanilla HTML page that's fine — browser defaults give headings a larger, bolder appearance. On any host that uses a CSS reset, **all heading levels collapse to body text size and weight**, and prompt markdown loses its visual hierarchy entirely.

This is what a researcher's `## Watch the clip` looks like on our platform (Tailwind v4 preflight resets headings to inherit):

![screenshot — H2 renders identical to the body paragraph below it](https://user-images.githubusercontent.com/PLACEHOLDER/markdown-h2-collapsed.png)

The markdown source is:

```markdown
## Watch the clip

Watch the entire video below, then answer the question. You can pause,
rewind, or rewatch as many times as you like before submitting your
response.
```

A researcher writing this expects "Watch the clip" to render as a heading. They get a paragraph that happens to start with capital letters.

## Why it's SCORE's job

A treatment file is portable across host platforms. The same SCORE study should look reasonable on any host. If three platforms ship the same study and three different headings render at three different sizes, the researcher has to debug each one individually. The styling of researcher-authored markdown is a property of "writing markdown", not a property of "running on platform X". SCORE owns the contract: "you give me markdown, I render it as readable formatted text".

This is the same logic that makes SCORE own button shapes, slider thumbs, and the media player controls.

## Why it's also the host's job, sometimes

Hosts have different visual languages. Our platform uses smaller headings than a Prolific deployment might. SCORE shouldn't dictate a fixed type scale — it should ship sensible defaults and let the host customize.

The precedent is `--score-primary` for accent colors. Extending that pattern to typography is the obvious shape.

## Suggested fix

In [`Markdown.tsx`](https://github.com/deliberation-lab/SCORE/blob/main/src/components/form/Markdown.tsx), pass a `components` map to `ReactMarkdown` that gives each element type **inline styles backed by CSS custom properties with sensible defaults**.

```tsx
return (
  <div
    id="markdown"
    className="max-w-xl"
    style={{
      fontSize: "var(--score-prompt-text-size, 1rem)",
      lineHeight: "var(--score-prompt-line-height, 1.5)",
    }}
  >
[...]

PR #390, opened 2026-05-20
UI polish: Markdown (#350 tier 2)

JamesPHoughton:
First component in tier 2 of the #350 polish sweep. Markdown renders every prompt body in every study, so the gains compound across all deployments.

## Visible changes

- **Links** — `:hover` (`--stagebook-link-hover` #1d4ed8), `:focus-visible` ring, `:visited` (`--stagebook-link-visited` #7c3aed). No focus indicator before this was a WCAG 2.4.7 fail. External links (`target=_blank`) get `rel="noopener noreferrer"` automatically. No external-link arrow glyph: adding chrome to researcher-authored text would visually rewrite what participants see.

- **`<pre>` keyboard-scrollable** — `tabIndex={0}` + focus-visible ring. Previously keyboard users couldn't horizontally scroll long code lines (WCAG 2.1.1).

- **h5 + h6 handlers** — previously absent; deeper headings collapsed to body-text size on hosts that strip the UA stylesheet. Added handlers + `--stagebook-prompt-h5-size` (1rem) / `-h6-size` (0.875rem) / matching weight variables.

- **Tables**: mobile overflow + zebra + row hover. Tables now live inside an `overflowX: auto` wrapper so wide tables on narrow viewports scroll independently. Body rows alternate background via `tr:nth-child(even)`; rows tint on hover. Matches GitHub / Notion consensus.

- **Blockquote refresh** — dropped redundant inner `maxWidth` (parent already caps at 36rem), tightened padding (taller-than-wide reads as quote vs. box), added muted `color` so quoted text visually steps back from body, bumped default border `#d1d5db` → `#9ca3af` so the rail actually reads against the page. Mirrored to [Display.tsx](packages/stagebook/src/components/elements/Display.tsx) (the inline duplication is intentional per #33). No italic — researcher quotes should render verbatim.

- **List markers muted** via `::marker` so bullet/number reads as structure, not weight-competing with body content.

- **Images** — `loading="lazy"` + `decoding="async"` so offscreen images in long prompts free bandwidth for above-the-fold content.

## Two latent bugs fixed along the way

- **Duplicate `id="markdown"`** — every Markdown instance shared the same id (invalid HTML on multi-Prompt stages). Switched to `useId()`-backed per-instance class. Same pattern as Button / Slider / ListSorter.

- **Fenced `<code>` showed per-line tint inside the chip in VS Code's webview.** The outer `<pre>` had our inline chip background, but the inner `<code>` had no inline style — so VS Code's default `code { background-color: var(--vscode-textPreformat-background) }` painted behind every text glyph. Inner code now carries `background: transparent` + `padding: 0` so it defeats host code styling. New regression test simulates the hostile rule.

## Architectural decision: scoped `<style>` per instance

State-dependent behavior (`:hover` / `:focus-visible` / `:visited` / `::marker` / `:nth-child`) isn't inline-expressible. One `<style>` block per instance, scoped to the `useId()`-derived class, carries those rules. Inline styles still own structural/dimensional properties; the scoped block only adds pseudo-class behavior. Hosts can't override the structural contract but Stagebook gains state behavior the inline-only philosophy couldn't express. Same pattern as Button / Slider / ListSorter / Markdown's siblings in tier 1.

## CSS variables added (to `styles.css` `:root`)

- `--stagebook-link-hover` (#1d4ed8)
- `--stagebook-link-visited` (#7c3aed)
- `--stagebook-prompt-h5-size` (1rem) / `-h6-size` (0.875rem)
- `--stagebook-prompt-h5-weight` / `-h6-weight` (both 600)
- Default bumped: `--stagebook-blockquote-border` #d1d5db → #9ca3af

Reuses existing `--stagebook-hover-bg` (table row hover), `--stagebook-text-muted` (list markers + blockquote text), `--stagebook-focus-ring` (link + pre focus rings), `--stagebook-bg-muted` (zebra stripe).

## Gallery demo expanded

Previous demo exercised h1-h3, lists, fenced code, blockquote, hr. Added: full h1-h6 hierarchy, GFM strikethrough, GFM task-list checkboxes, a 5-row GFM table, an inline image (data-URI placeholder). Also corrected the demo table copy — `submitButton` records a submit timestamp when named, so it's not in the same "no signal" bucket as `separator` / `audio` / `display`.

## Tests

47 Markdown CT pass (33 existing + 14 new):
- link hover / focus-visible
- external link rel attribute contract
- pre tabIndex + focus-visible
- h5 + h6 sized correctly
- table wrapped in overflow container
- table zebra striping
- table row hover
- blockquote muted color
- list marker color
- images lazy + decoding async
- two Markdown instances get unique classes (no duplicate id)
- fenced code inner-code inline background = transparent
- fenced code survives a hostile host code-styling rule

Full unit (1030) + CT (515) suites also pass.

## Test plan

- [x] `npm run test:ct -w stagebook -- --grep "Markdown"` (47/47 pass)
- [x] `npm test` (1030/1030 unit pass)
- [x] `npm run test:ct -w stagebook` (515/515 CT pass)
- [x] Visual smoke in the VS Code extension preview, gallery's expanded markdown stage
- [ ] CI green

### DIFF
--- a/apps/vscode/src/extension.ts
+++ b/apps/vscode/src/extension.ts
@@ -626,7 +626,7 @@ function getWebviewContent(
       --stagebook-link: #2563eb;
       --stagebook-code-bg: rgba(0, 0, 0, 0.06);
       --stagebook-code-font: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
-      --stagebook-blockquote-border: #d1d5db;
+      --stagebook-blockquote-border: #9ca3af;
       --stagebook-blockquote-bg: #f9fafb;
     }
     body {
--- a/examples/component-gallery/prompts/markdown_demo.prompt.md
+++ b/examples/component-gallery/prompts/markdown_demo.prompt.md
@@ -5,8 +5,18 @@ name: markdown_demo
 
 ## Headings, paragraphs, emphasis
 
-Paragraphs flow normally. **Bold**, *italic*, and `inline code` all
-work. So do [external links](https://github.com/deliberation-lab/stagebook).
+Paragraphs flow normally. **Bold**, *italic*, ~~strikethrough~~, and
+`inline code` all work. So do
+[external links](https://github.com/deliberation-lab/stagebook).
+
+### Heading hierarchy
+
+The full set (h1-h6). Researchers usually stop at h3, but the deeper
+levels are available for long-form instruction text.
+
+#### h4 — section
+##### h5 — sub-section
+###### h6 — minor heading
 
 ### Lists
 
@@ -19,6 +29,12 @@ work. So do [external links](https://github.com/deliberation-lab/stagebook).
 2. Ordered second
 3. Ordered third
 
+Task-list checkboxes work too:
+
+- [x] Saved consent
+- [x] Verified microphone
+- [ ] Completed practice round
+
 ### Code blocks
 
 ```yaml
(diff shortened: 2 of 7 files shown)


JamesPHoughton:
Subagent review: **SHIP WITH NITS** — all four addressed in 99b824f.

Highlights from the review:
- Pattern parity with Button / Slider / TextArea / Select / ListSorter is clean — useId + safeId regex, scoped <style>, inline-vs-CSS specificity split, border longhands all on-pattern.
- :focus-visible scoping correct; no other inline-vs-CSS specificity traps found.
- Fenced-code host-CSS diagnosis right; fix is complete for `background-color` (the actual VS Code rule). `font-family` and `line-height` already inherit from inline `<pre>`; `color` is undefended but no host known to attack that.
- Migration from `id="markdown"` → `useId()` class is clean — no stale consumers anywhere in the monorepo.
- Display.tsx blockquote mirror is byte-for-byte equivalent to Markdown.tsx's.

Addressed nits:
1. `apps/vscode/src/extension.ts` hardcoded the OLD `--stagebook-blockquote-border` in the webview :root — would have defeated the polish in the VS Code preview where we tested it. Updated to `#9ca3af`.
2. Stale comment in `Markdown.ct.tsx` referenced `#markdown ol > li::before` (selector no longer exists). Updated to `.stagebook-markdown-*`.
3. `target="_blank"` → `rel="noopener noreferrer"` contract had no real test. Extracted as `computeSafeRel()` (exported), covered by 5 unit tests in `Markdown.test.ts`. The existing CT now honestly describes itself as the baseline check.
4. Misleading comment on `<pre>` `tabIndex={0}` said it fires "whenever content overflows" but it's unconditional. Updated to match actual behavior.

Out-of-PR notes from the review (not addressed here):
- `color` could be added to the fenced-code defenses if a host ever turns up that paints text color through `code { color: ... }`. Not the reported bug — flag for later if it surfaces.

Issue #214, opened 2026-04-24
Migrate Markdown table styling into inline table/thead/tbody/tr/th/td handlers


JamesPHoughton:
## Problem

Markdown tables (`table`, `thead`, `tbody`, `tr`, `th`, `td`) have no handlers in [Markdown.tsx](packages/stagebook/src/components/form/Markdown.tsx)'s `components` map — they're styled entirely via rules in [styles.css:277-297](packages/stagebook/src/styles.css#L277-L297). The tension is self-flagged in that file's source comments (around #33):

> Intentionally NOT inlined into Markdown.tsx like headings/lists/etc. The table style surface is large (table + th + td + collapse behavior + cell borders) and inlining would roughly double Markdown.tsx. As a result: hosts that import this stylesheet get nice tables; hosts that don't get browser-default unstyled tables (functional but ugly).

With #211 setting the precedent that Markdown elements should self-style, tables are the biggest remaining gap. Researchers routinely use tables in prompts (condition matrices, payoff grids, comparison pairs).

## Proposed fix

Add `table:`, `thead:`, `tbody:`, `tr:`, `th:`, `td:` handlers in `Markdown.tsx`'s components map, each applying inline styles that mirror [styles.css:277-297](packages/stagebook/src/styles.css#L277-L297). Retire those rules from `styles.css` once migration lands so there's one source of truth. `thead` / `tbody` / `tr` may not need explicit styles if the default renders acceptably; verify and trim handlers we don't actually need.

## Out of scope

- Nothing related — scoped to the six table tags only.

## Related

- Surfaced by the post-#211 sweep.
- See #211 / #212 for the `<img>` precedent.

PR #392, opened 2026-05-20
UI polish: Separator (#350 tier 2)

JamesPHoughton:
Second component in tier 2 of the #350 polish sweep. Separator is a non-interactive horizontal rule, so most of the per-component checklist (hover / focus / keyboard / reduced-motion / touch-target / ARIA) doesn't apply — `<hr>` already carries implicit `role="separator"` with horizontal orientation. Polish here is internal cleanup + defensive fixes.

## Changes

- **Wrapper `<div>` dropped.** `<hr>` is already block-level, so the wrapper was a structurally pointless DOM node.

- **`border: "none"` shorthand → longhand (`borderStyle: "none"`).** React's inline-style diff has been observed to clear the `border` shorthand without re-emitting the longhand expansion when switching between styles — would surface here as a stray default border stripe overlapping the colored bar. Same defensive pattern as the #367 fixes in the form components.

- **Three ternaries → lookup map.** Three mutually-exclusive variant conditions collapse into a single `VARIANT_STYLES` table keyed by variant name. Slightly easier to extend if a future variant lands.

## Out of scope (intentional)

- Hover / focus / disabled / keyboard / reduced-motion / touch-target — `<hr>` isn't interactive.
- ARIA — `<hr>` already has implicit `role="separator"` + horizontal orientation.
- New `--stagebook-separator-color` token — current `--stagebook-text-faint` (thin/regular) and `--stagebook-text-muted` (thick) work; separator-specific variable is YAGNI without a real ask.
- Visual color tuning — would be a behavioral change without clear motivation.

## Tests

- Existing variant-height assertions updated to query `component` directly (when the root is the `<hr>` itself, `.locator("hr")` finds no nested hr and fails).
- **New** "renders <hr> directly (no wrapper div)" — guards against accidental reintroduction.
- **New** "uses border longhands, not the border shorthand" — reads the inline style attribute to confirm the #367 defense.
- **New** "legacy empty-string style resolves to regular" — locks in the prompt-file schema back-compat path.
- **New** "inline styles survive an aggressive host CSS reset" — mirror of the table inline-style test from issue #214; a Tailwind-style preflight `hr { background: transparent; height: 0 }` must not erase the visible rule.

All 8 Separator CT pass. 1035 unit + 519 of 520 CT pass (one unrelated `Timeline.ct.tsx:3085` flake passes on isolation re-run; the test is not touched here).

## Test plan

- [x] `npm run test:ct -w stagebook -- --grep "Separator"` (8/8 pass)
- [x] `npm test` (1035/1035 pass)
- [ ] CI green

### DIFF
--- a/packages/stagebook/src/components/form/Separator.ct.tsx
+++ b/packages/stagebook/src/components/form/Separator.ct.tsx
@@ -1,19 +1,99 @@
 import { test, expect } from "@playwright/experimental-ct-react";
 import { Separator } from "./Separator";
 
+// The Separator renders a single <hr> as its root element, so the
+// Playwright `mount` locator IS the <hr> — assert against `component`
+// directly rather than locator("hr") (which would search for a nested
+// hr inside the hr, which doesn't exist).
+
 test("renders regular separator by default", async ({ mount }) => {
   const component = await mount(<Separator />);
-  const hr = component.locator("hr");
-  await expect(hr).toBeVisible();
-  await expect(hr).toHaveCSS("height", "3px");
+  await expect(component).toBeVisible();
+  await expect(component).toHaveCSS("height", "3px");
 });
 
 test("renders thin separator", async ({ mount }) => {
   const component = await mount(<Separator style="thin" />);
-  await expect(component.locator("hr")).toHaveCSS("height", "1px");
+  await expect(component).toHaveCSS("height", "1px");
 });
 
 test("renders thick separator", async ({ mount }) => {
   const component = await mount(<Separator style="thick" />);
-  await expect(component.locator("hr")).toHaveCSS("height", "5px");
+  await expect(component).toHaveCSS("height", "5px");
+});
+
+// ----------- UI polish (#350 tier 2) -----------
+
+test("legacy empty-string style resolves to regular (back-compat)", async ({
+  mount,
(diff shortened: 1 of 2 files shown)


JamesPHoughton:
Subagent review: **SHIP**.

Highlights:
- Scope correctly skips hover/focus/keyboard/reduced-motion/touch-target for a non-interactive `<hr>`; the four items in scope (drop wrapper, border longhand, lookup-map refactor, host-reset regression) are the right ones.
- `Exclude<SeparatorProps["style"], undefined | "">` correctly narrows the prop union; the `""` → `"regular"` collapse is type-safe and local.
- `borderStyle: "none"` is the right longhand — functionally equivalent to `border: "none"` for rendering but stable across React style diffs.
- Tests: variant heights assert against `component` (the `<hr>` root), the "no wrapper div" tag-name check catches reintroduction, host-reset test mirrors the #214/#215 pattern.

Out-of-PR observation (not a blocker):
- `Markdown.tsx:189` still uses `border: 0` shorthand for its `<hr>` rendering, slightly inconsistent with the new Separator pattern. Markdown's `<hr>` separately sets the longhand `borderTop*` so it's defensible, but worth aligning on a sweep cleanup later.

PR #449, opened 2026-05-28
Add stagebook/dispatch: urn + uniform-random + pluggable contract harness

### LINKED ISSUE #448: Add urn-model dispatcher + pluggable contract harness
## Motivation

The dispatcher that assigns participants to treatments currently lives in deliberation-lab/server/preFlight/dispatch.js as a single payoff-weighted DFS optimizer. But **most experiments are not Bayesian optimization** — they're "randomize across treatments with target sample sizes." The current dispatcher's BO framing is a poor default for ordinary experimental work, and its randomization story is hard to defend in a methods section without a careful explanation of acquisition functions.

Three distinct sampling models cover the use cases we actually encounter:

| Model | What you provide | What you get | When to use |
|---|---|---|---|
| **Uniform random** | nothing | Each slot is an independent uniform draw | Quick prototypes, no balance requirement |
| **Urn** | target counts | Exactly your target counts (sampling without replacement) | Most randomized experiments — *the default* |
| **Local penalization** | acquisition values + penalization matrix | One batch step of a batched-BO loop | Studies running a researcher-managed BO surrogate |

This issue adds the **first two** to stagebook: `uniform-random` and `urn`. The existing BO-style dispatcher stays in deliberation-lab and gets formally renamed `local-penalization` (its real algorithmic identity per Gonzalez et al. 2016, *Batch Bayesian Optimization via Local Penalization*), with a docstring clarifying it implements one batch step of an iterated BO loop where the researcher acts as the surrogate model between batches.

This split lives in stagebook because:
- The assignment algorithm is part of the **study definition** — papers should cite the algorithm by name from the open-source DSL package
- Reviewers can audit the algorithm in the same npm package referenced by the study DSL
- Composes naturally with stagebook's existing eligibility-condition system (`compare`, `getReference`)
- Portability: re-running the algorithm on collected data for post-hoc analyses (e.g. IPW from Monte-Carlo assignment probabilities) is straightforward when it's a pure-function npm import

## Scope

New `src/dispatch/` module:

```
stagebook/src/dispatch/
├── types.ts                    Treatment, EligibilityTable, Assignment, DispatchResult,
│                               dispatcher config discriminated unions
├── extractConditionKeys.ts     pure: Treatment[] → Set<string>
├── makeEligibilityTable.ts     pure: (playerIds, treatments, playerData) → EligibilityTable
├── uniformRandom.ts            dispatcher #1 — iid sampling with replacement
├── urnRandomization.ts         dispatcher #2 — sampling without replacement against counts
├── validateDispatcherConfig.ts pure: validates parameter shapes + semantic invariants
├── contract.ts                 generic contract harness (parameterized over dispatcher factory)
└── index.ts                    re-exports
```

The existing `local-penalization` dispatcher stays in deliberation-lab. All three (the two new ones + local-penalization) run against the generic contract harness exported from stagebook.

### Dispatcher selection in the batch config

The batch config gets a `dispatcher` field that's a discriminated union:

```yaml
# Simplest — no balance guarantees, just iid coin-flip per slot
dispatcher:
  type: uniform-random

# Default for most experiments — sampling without replacement
dispatcher:
  type: urn
  counts: [2, 2, 2, ...]
  # decrements omitted → defaults to identity (decrement self by 1)

# Cross-treatment relationships (pilot_3-style)
dispatcher:
  type: urn
  counts:
    from: ./counts.json        # file reference for large studies
  decrements:
    from: ./decrements.json    # square N×N matrix

# Opt-in for BO loops
dispatcher:
  type: local-penalization
  payoffs: [...]               # acquisition function values from researcher's surrogate
  knockdowns: ...              # within-batch diversification matrix
```

**Backward compatibility**: legacy configs with top-level `payoffs` + `knockdowns` (no `dispatcher` field) get rewritten internally:
- `payoffs: "equal"` + `knockdowns: "none"` → `dispatcher: { type: "uniform-random" }` (makes explicit what the algorithm actually was)
- `payoffs: <array>` + `knockdowns: anything` → `dispatcher: { type: "local-penalization", payoffs, knockdowns }`

Both rewrites emit a deprecation warning encouraging migration to the explicit form.

### Urn dispatcher parameter shape

Minimized to just what the model actually needs:

```ts
type UrnDispatcherConfig = {
  type: "urn";
  counts: number[] | { from: string };          // required, non-negative integers
  decrements?: number[][] | { from: string };   // optional, square matrix
                                                // default: identity (decrement self by 1)
};
```

Scalar / per-treatment-list forms for `decrements` are deliberately excluded — both are redundant with adjusting `counts` in the integer-subtractive case (`counts = c, decrements = d` is equivalent to `counts = ceil(c/d), decrements = 1`). The only configuration that *isn't* expressible via counts alone is cross-treatment relationships, which need a matrix.

Validation:
- `counts[i]` non-negative integer
- `decrements[i][j]` non-negative integer
- `decrements[i][j] <= counts[j]` (caught at config-time; can't subtract more balls than exist initially)
- Matrix is square, dimensions match `counts.length`
- Mid-dispatch underflow (after multiple picks) clamps to zero with optional `info()` log

### Public API sketch (host-side per-tick flow)

```ts
const keys = stagebook.extractConditionKeys(treatments);
const playerData = buildPlayerDataSnapshot(empiricaPlayers, keys);  // host code
const table = stagebook.makeEligibilityTable({ playerIds, treatments, playerData });
const result = stagebook.urnRandomization({
  playerIds,
  treatments,
  counts: remainingCounts,
  decrements,
  eligibility: table,
  rng,
});
remainingCounts = result.remainingCounts;  // host persists across ticks
```

Key design constraints (arrived at via deliberation-lab#268 discussion):
- **Eligibility is pre-computed** before dispatch — decouples the dispatcher from reference resolution
- **Dispatcher sees only IDs + structural facts** (positionCounts, counts, decrements, eligibility table); no PlayerView, no references, no stagebook conventions reach the algorithm itself
- **Counts are host-persistent state** — dispatcher is pure, host carries `remainingCounts` between ticks
- **Decrement matrix encodes researcher-owned discrete-space locality** — since treatments are labeled cells not geometric points, the researcher specifies which treatments are "near" each other for diversification purposes (cf. Gonzalez et al. 2016 for the continuous-space analogue this generalizes from)

### Generic contract harness

The existing dispatcher contract harness in deliberation-lab (`dispatch.contract.test.js`) is already parameterized over a dispatcher factory but has BO-specific terminology in its scenario generator. Move it to stagebook and generalize:
- Scenario shrinks to `{ players, treatments }`
- Each registered dispatcher's factory generates algorithm-specific params from a seeded rng, returns `{ dispatch, params }`
- `params` is included in failure dumps so debug output stays informative per-algorithm

All three dispatchers (uniform-random + urn in stagebook, local-penalization in deliberation-lab) run against the same 10 structural invariants.

## Statistical-properties test suite (8 tests at α=1e-4)

In addition to the structural contract, both new dispatchers need a randomization-properties suite — the receipts a reviewer would want for a paper's randomization claim. Each test uses chi-square goodness-of-fit or binomial z-tests with α=1e-4, M chosen so binomial SE detects ≥2% absolute deviations:

1. **Seeded determinism** — same seed + inputs ⇒ identical output. Reproducibility foundation.
2. **Marginal target rate** — over many ticks, each treatment is used at its target count. *Central claim of urn randomization.*
3. **Position uniformity** — symmetric players are assigned to position 0 vs. 1 of their group with equal frequency.
4. **Equal opportunity within eligibility class** — players with identical eligibility have equal probability of being the leftover (or being assigned).
5. **Arrival-order independence** — input ordering does not shift the aggregate (player → position) distribution.
6. **Irrelevant-attribute independence** — a player attribute that no condition references (but happens to correlate with input order) does not predict assignment. *Headline test for reviewer defensibility.*
7. **Pairwise co-assignment independence** — for any pair of players, P(co-assigned) matches the constraint-implied baseline (no spurious clumping).
8. **Variant balance within label** — for matrix-decrement designs, the variants within a label are used uniformly when the label is repeatedly selected.

See [deliberation-lab/server/src/preFlight/dispatch.randomization.test.js](https://github.com/deliberation-lab/deliberation-lab/blob/main/server/src/preFlight/dispatch.randomization.test.js) (from deliberation-lab#268) for the reference implementation against the existing BO-style dispatcher.

## Non-scope

Explicitly deferred:
- **Monte-Carlo assignment-probability wrapper for IPW** — the pure-function design makes this a trivial later addition, but the surrounding question of *which players belong in a dispatch batch* has nuances that warrant their own design pass
- **Migrating `local-penalization` from deliberation-lab to stagebook** — leave it where it is; the new dispatchers are additive
- **Refactoring `local-penalization` to use the same pre-computed `EligibilityTable`** — possible follow-up cleanup PR after the new dispatchers land
- **Global state in conditions** (e.g., a `lobby.X` reference form for "number of waiting players") — the design accommodates this as a future extension but doesn't need it for current work

## Related

- deliberation-lab#267 — matrix-knockdown perf + marginals harness at n=420 (merged)
- deliberation-lab#268 — Fisher-Yates shuffle fix + 8-test randomization properties suite for `local-penalization` dispatcher (merged); the 8 tests above mirror what landed there

JamesPHoughton:
## Summary

Adds the new `stagebook/dispatch` module from #448:

- **`uniform-random`** — trivial iid-by-treatment dispatcher; explicit name for what `payoffs: "equal", knockdowns: "none"` actually was.
- **`urn`** — sampling-without-replacement against per-treatment target counts, with an optional N×N integer decrement matrix for cross-treatment locality. Pure function; host carries `remainingCounts` across ticks.
- **Generic contract harness** (`stagebook/dispatch/contract`) — the 10 structural invariants from deliberation-lab's `dispatch.contract.test.js`, scenario-shrunk to `{ players, treatments }` and parameterized over a dispatcher factory. deliberation-lab will register its existing `local-penalization` against this same gauntlet in a follow-up PR.
- **Pre-computed eligibility** — `extractConditionKeys` + `makeEligibilityTable` decouple the dispatcher from reference resolution per the issue's design constraints; the algorithms themselves see only IDs + structural facts.
- **Config validator** — `validateDispatcherConfig` checks the discriminated-union shape; recognizes `local-penalization` without deep-validating its shape (deliberation-lab owns that).

Package exports added: `stagebook/dispatch` (main API, vitest-free) and `stagebook/dispatch/contract` (harness — depends on vitest).

### Test coverage

- **Contract gauntlet (20 tests):** 10 invariants × 2 dispatchers × 200 random scenarios each. Invariant #9 now snapshots `params` too (so e.g. urn's `counts` input immutability is asserted as part of the contract).
- **Statistical-properties suite (15 tests, α=1e-4):** mirrors the 8-test pattern from deliberation-lab#268 — seeded determinism, marginal target rate, position uniformity, equal-opportunity-within-class, arrival-order independence, irrelevant-attribute independence, pairwise co-assignment independence; plus the matrix-decrement variant-balance test for urn (uniform-random has no decrement matrix, so this one is urn-only).
- **Unit tests (29 tests):** `extractConditionKeys`, `makeEligibilityTable`, `validateDispatcherConfig`.

64 new tests; all 1309 stagebook tests pass.

### Out of scope (per the issue's Non-scope section)

- Migrating `local-penalization` itself into stagebook (stays in deliberation-lab).
- Batch-config schema integration + backward-compat shim that rewrites legacy `payoffs`/`knockdowns` into the new `dispatcher: { type, ... }` shape — that lives in deliberation-lab.
- Monte-Carlo assignment-probability wrapper for IPW.

### Notes for reviewers

- Mid-dispatch underflow on `decrements[i][j]` clamps to zero (matches the issue body). The issue mentions an optional `info()` log; the implementation skips that — hosts can detect the same condition by diffing input vs. output counts and surface it however their observability stack prefers.
- The greedy fill in `tryFillTreatment` shuffles positions before assigning so strict-eligibility positions don't always win the race with unconstrained positions in the same treatment.

Closes #448.

## Test plan

- [x] `npm run lint -w stagebook` clean
- [x] `npm test -w stagebook` — 1309/1309 pass (new 64 + existing 1245)
- [x] `npm run build -w stagebook` succeeds; new bundle entries `dist/dispatch/index.{js,cjs,d.ts}` and `dist/dispatch/contract.{js,cjs,d.ts}` produced
- [x] Both dispatchers pass the contract gauntlet (10 invariants × 200 scenarios)
- [x] All 15 statistical properties pass at α=1e-4

### DIFF
--- a/packages/stagebook/package.json
+++ b/packages/stagebook/package.json
@@ -38,6 +38,26 @@
         "types": "./dist/validate/index.d.cts",
         "default": "./dist/validate/index.cjs"
       }
+    },
+    "./dispatch": {
+      "import": {
+        "types": "./dist/dispatch/index.d.ts",
+        "default": "./dist/dispatch/index.js"
+      },
+      "require": {
+        "types": "./dist/dispatch/index.d.cts",
+        "default": "./dist/dispatch/index.cjs"
+      }
+    },
+    "./dispatch/contract": {
+      "import": {
+        "types": "./dist/dispatch/contract.d.ts",
+        "default": "./dist/dispatch/contract.js"
+      },
+      "require": {
+        "types": "./dist/dispatch/contract.d.cts",
+        "default": "./dist/dispatch/contract.cjs"
+      }
     }
   },
   "files": [
--- a/packages/stagebook/src/dispatch/contract.test.ts
+++ b/packages/stagebook/src/dispatch/contract.test.ts
@@ -1,0 +1,75 @@
+// Run the generic dispatcher-contract gauntlet (10 structural
+// invariants) against each dispatcher stagebook ships. Adding a new
+// dispatcher means appending one `runContractSuite(...)` call; the
+// same invariants run automatically.
+
+import { buildEligibilityForScenario, runContractSuite } from "./contract.js";
+import { uniformRandom } from "./uniformRandom.js";
+import { urnRandomization } from "./urnRandomization.js";
+
+runContractSuite("uniform-random", ({ scenario, rng }) => {
+  const eligibility = buildEligibilityForScenario(scenario);
+  return {
+    params: {},
+    dispatch: () =>
+      uniformRandom({
+        playerIds: scenario.players.map((p) => p.id),
+        treatments: scenario.treatments,
+        eligibility,
+        rng,
+      }),
+  };
+});
+
+runContractSuite("urn", ({ scenario, rng }) => {
+  const eligibility = buildEligibilityForScenario(scenario);
+  // Random per-scenario counts in [0, 5]. Some zeros so the dispatcher
+  // exercises the "no balls left for this treatment" branch.
+  const counts = scenario.treatments.map(() => Math.floor(rng() * 6));
+  // Random decrement matrix: either identity (50%) or a sparse matrix
+  // with integer entries small enough to avoid validation rejections.
+  let decrements: number[][] | undefined;
+  if (rng() < 0.5) {
+    decrements = undefined; // identity default
+  } else {
+    decrements = buildSafeDecrementMatrix(counts, rng);
+  }
+  return {
+    params: { counts, decrements },
+    dispatch: () =>
+      urnRandomization({
+        playerIds: scenario.players.map((p) => p.id),
+        treatments: scenario.treatments,
+        counts,
+        decrements,
+        eligibility,
+        rng,
+      }),
+  };
+});
+
+/** Build a random decrement matrix where each `decrements[i][j] ≤ counts[j]`
+ *  so the dispatcher doesn't immediately clamp to zero on the first use.
+ *  Diagonal entries are at least 1 so a treatment's own counter
+ *  actually decreases when the treatment is used. */
+function buildSafeDecrementMatrix(
+  counts: number[],
+  rng: () => number,
+): number[][] {
+  const n = counts.length;
+  const m: number[][] = [];
+  for (let i = 0; i < n; i += 1) {
+    const row: number[] = [];
+    for (let j = 0; j < n; j += 1) {
+      if (i === j) {
+        // Self-decrement at least 1 (when there's a ball to take).
+        row.push(counts[j] > 0 ? 1 : 0);
+      } else {
+        // Off-diagonal in [0, min(1, counts[j])] — usually 0, sometimes 1.
+        row.push(rng() < 0.25 && counts[j] > 0 ? 1 : 0);
+      }
+    }
+    m.push(row);
+  }
+  return m;
+}
--- a/packages/stagebook/src/dispatch/contract.ts
+++ b/packages/stagebook/src/dispatch/contract.ts
@@ -1,0 +1,485 @@
+// mulberry32 below uses bit ops by design (small fast PRNG, intentional).
+
+// Generic dispatcher contract harness (#448).
+//
+// Every dispatcher implementation — current and future, stagebook's
+// own (`uniform-random`, `urn`) and external (`local-penalization` in
+// deliberation-lab) — MUST satisfy the structural invariants pinned
+// here. The harness is parameterized over a *factory* so registering a
+// new algorithm is a matter of supplying its scenario-to-dispatcher
+// adapter and re-running the gauntlet.
+//
+// Scope: structural invariants only. Algorithm-specific statistical
+// claims (marginal target rate, irrelevant-attribute independence, …)
+// live in algorithm-specific test files because each algorithm makes a
+// different randomization claim worth a paper-defensible receipt.
+//
+// History: this harness is the generalization of
+// `dispatch.contract.test.js` in deliberation-lab (was tightly coupled
+// to payoffs+knockdowns shape). The scenario was shrunk to just
+// `{ players, treatments }`; algorithm-specific params come from the
+// registered factory (which can also generate them from the seeded
+// rng so the gauntlet's full input space remains deterministic).
+
+import { describe, test, expect } from "vitest";
+import type { DispatchResult, Treatment } from "./types.js";
+import { makeEligibilityTable } from "./makeEligibilityTable.js";
+import { evaluateConditions } from "../utils/evaluateConditions.js";
+import {
+  getNestedValueByPath,
+  getReferenceKeyAndPath,
+} from "../utils/reference.js";
+
+/** Mulberry32 — small, fast, fully deterministic PRNG. The bit ops are
+ *  intentional (PRNG); we don't need cryptographic quality, we need
+ *  reproducibility. Returned as a function so callers can pass it
+ *  anywhere a `() => number` rng is expected. */
+export function mulberry32(seed: number): () => number {
+  let s = seed >>> 0;
+  return () => {
+    s = (s + 0x6d2b79f5) >>> 0;
+    let t = s;
+    t = Math.imul(t ^ (t >>> 15), t | 1);
+    t ^= t + Math.imul(t ^ (t >>> 7), t | 61);
+    return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
+  };
+}
+
+/** One harness scenario — players + treatments, nothing else. Each
+ *  registered dispatcher's factory builds its own algorithm-specific
+ *  params on top. */
+export interface ContractScenario {
+  players: {
+    id: string;
+    /** Storage-key indexed snapshot, populated for the keys returned
+     *  by `extractConditionKeys` over `treatments`. */
+    data: Record<string, Record<string, unknown>>;
+  }[];
+  treatments: Treatment[];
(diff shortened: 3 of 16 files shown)


JamesPHoughton:
Reviewed by a separate agent. **Recommend merge** — faithful implementation of the design in #448 with two intentional refinements worth highlighting.

## What I verified

- Re-ran the dispatch suite locally: **64/64 pass in 1.07s**
- API surface matches #448 (types, exports, configs, validation rules)
- Cross-checked the statistical tests against deliberation-lab#268's reference implementation
- Read all 6 source files + 4 test files + package.json/tsup config

## Strong points worth keeping

1. **Position-shuffle in `tryFillTreatment`** — fills positions in randomized order rather than 0→playerCount-1. Eliminates a positional-bias failure mode that wasn't explicitly called out in the issue but is a real improvement over deliberation-lab's existing dispatcher. Smart defense for invariant #4 + the irrelevant-attribute test.

2. **Invariant #9 now snapshots `params` too** — important contract because hosts thread `remainingCounts` across ticks; if the dispatcher mutated it, the host's persisted state would silently corrupt.

3. **Independent `playerSatisfies` in the harness** — invariant #4 checks the dispatcher's eligibility against an independent re-resolution via `evaluateConditions` directly. Means the eligibility table can't mark its own homework.

4. **Two-subpath export** — `stagebook/dispatch` (runtime) vs. `stagebook/dispatch/contract` (test harness) keeps vitest out of runtime consumers' bundles.

5. **Defensive weighted-sample fallback in `urnRandomization`** — `let treatmentIdx = pool[pool.length - 1]` handles the floating-point edge case where `target === totalWeight` and the cumulative scan never triggers.

6. **Snapshot timing in the contract harness** — captures `treatmentsBefore` / `playerDataBefore` / `paramsBefore` AFTER factory construction but BEFORE dispatch, so the invariant check sees what dispatch was actually given (rather than failing on a legitimate factory-time param generation).

## Deliberate departures from the issue body (both reasonable)

- **No `info()` log on mid-dispatch underflow clamping.** Host can detect by diffing input vs. output counts. Keeps the dispatcher silent; observability lives at the boundary.
- **`uniform-random` rejects extra keys at validation time** (a check not in the spec). Catches "I put counts here by mistake" errors with a clear remediation message. Worth keeping.

## Minor follow-ups (not blockers)

1. `isDispatchExhausted` helper from the design discussion isn't here — trivial one-liner that deliberation-lab#269 will need. Can land in #269's PR.
2. `mulberry32` is exported from `stagebook/dispatch/contract` (which pulls vitest), not from the runtime subpath. Deliberation-lab's integration (deliberation-lab#270) will keep its own 8-line copy in `math.js` rather than pulling vitest at runtime. Worth noting in the integration PR.
3. `evaluateConditions` is called via `Parameters<typeof evaluateConditions>[0]` in contract.ts — works, but slightly opaque. Future cleanup if that function's signature ever changes.

Ready to merge from a code-quality standpoint.

PR #488, opened 2026-06-29
fix(prompts): make the colon the numeric-mode signal for multipleChoice (#289)

### LINKED ISSUE #289: multipleChoice numeric/text auto-detection false-positives on comprehension checks
## The problem

After #282, multipleChoice options must be either all-numeric (`<number>(: <label>)?` on every line) or all-text (no line begins with a number). Mixed → preflight error:

> multipleChoice prompt mixes numeric and text response options. Either give every option a `<number>: <label>` prefix or remove all numeric prefixes.

Implementation: [promptFile.ts:413-433](https://github.com/deliberation-lab/stagebook/blob/main/packages/stagebook/src/schemas/promptFile.ts#L413-L433), with the numeric-detection helper at [promptFile.ts:195-202](https://github.com/deliberation-lab/stagebook/blob/main/packages/stagebook/src/schemas/promptFile.ts#L195-L202).

This false-positives on a real research pattern: **comprehension checks / quizzes** where numeric-looking option labels are identifiers, not scale points, and one of the answers is naturally non-numeric.

## Concrete case

A listener-instructions comprehension check from a study I'm migrating ([t_tapping/.../listener_comp_check_num_speakers.prompt.md](https://github.com/Talk-Lab/backchannel-manipulation/blob/main/stagebook/study/t_tapping/game/practice/listener_comp_check_num_speakers.prompt.md)):

```yaml
---
type: multipleChoice
---

## How many different people will tell you a story?

---

- 1
- 2
- 3
- 4
- It Varies
```

Used as a gate — the next stage shows only when `prompt.<name> == \"2\"`. The labels happen to look like Likert points but are not: there's no \"score,\" no averaging intent, no monotonic semantics. \"It Varies\" is a real first-class option (and is the foil the author wants the participant to *not* pick).

This shape recurs anywhere a study uses MCQ to gate on a discrete fact: \"How many rounds?\" with an \"I'm not sure\" foil; \"On which day?\" with a \"More than one\" option; demographics-style \"How many children do you have?\" with \"Prefer not to say\"; etc.

## Considerations

- **The rule's motivation is sound for Likert.** When every option carries a scale point, mixing — \"1: Strongly disagree, 2, 3, 4, 5: Strongly agree, Refused\" — is the case where downstream analysis silently breaks (what's the score for \"Refused\"?). #282 was right to fence that off, especially given the multi-select-numeric averaging hole called out in the inline comment.
- **The rule's heuristic doesn't distinguish intent.** It infers \"this prompt is on a numeric scale\" purely from the surface form of the labels. Comprehension checks where the labels happen to be small integers look identical to Likert prompts at the schema level.
- **There is currently no opt-out.** Researchers who hit this can only work around it by making every label non-parseable as a number — e.g. `1 person / 2 people / 3 people / 4 people / It varies` — and then updating any condition values that reference the label (`value: \"2\"` → `value: \"2 people\"`). Workable but viral: it changes the user-visible label text and the YAML wiring, both for a purely cosmetic reason.
- **Silent demotion would also have a downside.** A rule like \"if any option doesn't parse as numeric, drop the whole prompt to text mode\" would unblock comprehension checks but would also silently demote a researcher's intended Likert-with-Refused-option to text — losing the numeric semantics they actually wanted. Whichever direction the heuristic defaults, *someone's* intent gets misread.
- **An explicit mode flag avoids the inference question.** A frontmatter switch (some name; `mode:` / `numeric:` / etc.) would let a researcher state intent directly and turn auto-detection off. Adds a knob; resolves the ambiguity.
- **Other escape hatches exist.** Per-option markers (a sentinel that disables the numeric parse for one line), or recognising quoted-string options as forced-text, would also work. Each has its own ergonomics tradeoff.

Filing for visibility and to gather thoughts; I don't have a strong view on which direction to go.

## Workaround in the meantime

Replace bare numeric labels with non-parseable ones (`1 person`, `2 people`, …, `It varies`) and update any `comparator: equals / value:` conditions that referenced the old label.

JamesPHoughton:
Fixes #289.

## Problem

`multipleChoice` numeric mode (#282) inferred "this is a numeric scale" from the **surface form** of the labels: any option beginning with a number counted as numeric. A comprehension check / quiz whose options are small integers plus a non-numeric foil — e.g.

```
- 1
- 2
- 3
- 4
- It Varies
```

therefore tripped the *mixes numeric and text* preflight error, even though the labels are identifiers used to gate on a discrete fact (`prompt.<name> == "2"`), not scale points. The only workaround was to make every label non-parseable (`1 person`, `2 people`, …) and rewire any `value:` conditions — viral and cosmetic.

## Change

Make the explicit `- <number>: <label>` **colon form** the numeric signal (the design the issue author proposed — no new frontmatter knob). A bare `- 1` is now text (label `"1"`), shorthand for `- 1: 1`, never a scale point:

- **all** options explicit-numeric → numeric mode (`responsePoints` populated)
- **any** bare/text option mixed with explicit-numeric → mixing error (unchanged)
- otherwise (including all-bare) → text mode

So comprehension checks stay text and `value: "2"` conditions keep matching the bare label. To opt a scale into numeric semantics, label every option with a colon — give label-less middle points a trailing colon (`- 2:`), which falls back to the stringified number. The endpoints-only Likert pattern migrates cleanly:

```
- 1: Strongly disagree
- 2:
- 3:
- 4:
- 5: Strongly agree
```

## ⚠️ Breaking behavior change

A `multipleChoice` that used **bare** numbers (`- 1` / `- 2` / `- 3`) to get numeric mode is now **text** mode (`responsePoints` empty; labels `"1".."3"`). The saved value is still the bare label, so `value:` conditions are unaffected — only the numeric `responsePoints` (averaging semantics) are dropped for the bare form.

No example or test in the tree relied on bare-number numeric mode, and numeric mode (#282) is recent, so blast radius is minimal. **Recommend cutting as a minor bump (0.19.0)** with this note rather than a patch.

## Tests (red/green TDD)

- Flipped the old `bare numbers → numeric` test in place to assert text mode (regression lock).
- Added: comprehension check (`1/2/3/4/It Varies` → text, not error); endpoints-only Likert via trailing colons → numeric; colon-bearing **text** label stays text (`- Maybe: not sure`); single bare numeric → text; explicit-numeric mixed with a bare number errors in **both** orderings.

## Reviews

Ran correctness / coverage / security subagent reviews against the diff before opening:
- **Correctness**: change is correct; sliders unaffected (they call `parseNumericResponseLine` directly); fixed one stale `responsePoints` JSDoc that still said "or bare `- 1`".
- **Coverage**: added the colon-in-text-label and both mixing-order tests it flagged.
- **Security**: no vuln / no ReDoS (only `indexOf`/`slice`/`Number`); the only finding is the data-semantics migration concern captured in the breaking-change note above.

Docs: added a "Numeric scales" subsection to `docs/researcher/prompts.md` documenting the colon opt-in.

Full local vitest (1570 lib + viewer + vscode) green; Playwright CT green (one known unrelated webkit `CheckboxGroup` focus-ring flake, passed on re-run).

### DIFF
--- a/docs/researcher/prompts.md
+++ b/docs/researcher/prompts.md
@@ -102,6 +102,32 @@ Each option on its own line, prefixed with `- `:
 - Option C
 ```
 
+The saved value is the chosen option's text.
+
+**Numeric scales (#282).** To attach numeric scale points (averaged downstream), label every option in the explicit `- <number>: <label>` form. The colon is what turns the prompt numeric — it's an opt-in:
+
+```markdown
+- 1: Strongly disagree
+- 2: Disagree
+- 3: Neutral
+- 4: Agree
+- 5: Strongly agree
+```
+
+Each numeric value must be unique, and numeric mode is single-select only.
+
+A **bare** option that happens to look like a number (`- 2`, no colon) is **text**, with the number as its label — never a scale point (#289). So a comprehension check or quiz whose options are small integers plus a non-numeric foil stays in text mode, and `value:` conditions match the bare label:
+
+```markdown
+- 1
+- 2
+- 3
+- 4
+- It Varies
+```
+
+Mixing explicit `- <number>: <label>` options with bare/text options in the same prompt is a preflight error — make every option numeric (add a trailing colon to label-less points, e.g. `- 2:`) or none.
+
 ### Open Response
 
 Placeholder text prefixed with `> `:
--- a/packages/stagebook/src/schemas/promptFile.test.ts
+++ b/packages/stagebook/src/schemas/promptFile.test.ts
@@ -722,7 +722,10 @@ How much do you agree?
       }
     });
 
-    test("bare numbers are accepted; label defaults to stringified number", () => {
+    test("bare numbers are text mode; numeric mode requires an explicit colon (#289)", () => {
+      // The colon is the numeric-mode signal. A bare `- 1` is shorthand for
+      // `- 1: 1` (text label "1"), never a numeric scale point. Authors who
+      // want numeric semantics (responsePoints) must write `- 1: <label>`.
       const markdown = `---
 name: scale
 type: multipleChoice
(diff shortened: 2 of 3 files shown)


JamesPHoughton:
@codex review

JamesPHoughton:
@codex review

Issue #502, opened 2026-07-06
Publish the preview/viewer surface as a stagebook/viewer subpath export


JamesPHoughton:
## Motivation

The reference viewer's reusable harness (`apps/viewer/src/preview.ts`) now has two real consumers with a third incoming:

- the standalone GitHub-Pages viewer app,
- the VS Code extension (consumes it via workspace link — `"stagebook-viewer": "*"`),
- **TalkBench's Design tool Preview lens**, which needs to embed a participant-perspective preview over study files served from its own backend.

But `stagebook-viewer` is `private: true`, so the surface is unreachable from npm — external consumers must vendor or git-dep the monorepo.

## Proposal: a `stagebook/viewer` subpath of the existing published package — not a separate package

- **Version skew is the killer for a separate package.** stagebook is 0.x with breaking minors (reference position-selector prefixes, `urlParams`→`entryUrl`, the `attributes` merge), and hosts like TalkBench pin an exact stagebook version per study. A second package would need lockstep releases forever; a subpath means one version to pin, and the viewer can never disagree with the schema/templates it renders.
- **It's nearly free on weight.** The package's runtime deps already include `react-markdown`, `remark-gfm`, and `@hello-pangea/dnd`; react/react-dom remain optional peers exactly as for `./components`.
- **The Pages app dogfoods the API**: `apps/viewer` becomes a thin shell over the published export.

## Proposed surface

Mostly promoting what `preview.ts` already exports privately:

- **`<PreviewHost>`** — batteries-included harness: parsed treatment file + two callbacks (`getTextContent` async, `getAssetURL` sync, `loadImport` for `imports:`), owns expansion (`fillTemplates` + import resolution), the **mock state store** (simulated responses, resettable), and the dev chrome: treatment picker, intro-sequence picker, stage navigation, position/player selector, timeline scrubber.
- **`<Viewer>`** — the controlled inner component for hosts that bring their own chrome (expanded treatment + selection state/callbacks + the same content fns).
- **Introspection utilities** as stable API: `flattenSteps`, `extractStageReferences`, `extractTimeBreakpoints`, `createViewerStateStore` — useful for custom harnesses and structural tooling beyond preview.
- **Content-fn helpers**: `createStaticContentFns(files)` (tests/fixtures); keep `createUrlContentFns` (raw-GitHub loader) as an optional convenience for the standalone app.

## Boundary (what stays out)

[...]

Issue #61, opened 2026-04-13
Restructure repo as monorepo to support viewer app


JamesPHoughton:
## Context

With the SCORE → Stagebook rename complete, the next step is a viewer app that renders Stagebook studies for researchers and study designers (#57 established the public-facing identity). The viewer will live in this repo alongside the component library, so we need a monorepo structure that keeps these as separate concerns with independent build/deploy pipelines.

## Proposed structure

```
stagebook/
├── packages/
│   └── stagebook/              # npm library (moved from root)
│       ├── src/
│       ├── package.json        # @deliberation-lab/stagebook
│       ├── tsconfig.json
│       ├── tsup.config.ts
│       ├── vitest.config.ts
│       └── playwright-ct.config.ts
├── apps/
│   └── viewer/                 # Vite SPA (new)
│       ├── src/
│       ├── package.json
│       └── vite.config.ts
├── docs/                       # stays at root (repo-level)
├── package.json                # workspace root (not published)
├── pnpm-workspace.yaml         # or npm workspaces
├── CLAUDE.md
└── .github/
    └── workflows/
        ├── ci.yml              # PR checks: lint, test, build (all packages)
        ├── publish-library.yml # on GitHub Release → npm publish
        └── deploy-viewer.yml   # on push to main → GitHub Pages
```

## Tasks

- [ ] Set up workspace configuration (npm workspaces or pnpm)
- [ ] Move library code from root into `packages/stagebook/`
- [ ] Move shared config (`.prettierrc`, base `tsconfig.json`) to root, have packages extend them
- [ ] Create `apps/viewer/` scaffold (empty Vite + React app that imports from the library)
- [ ] Verify workspace linking: `apps/viewer` can `import { Stage } from "@deliberation-lab/stagebook/components"`
- [ ] Update CI workflow to run lint/test/build across all workspaces
- [ ] Add `publish-library.yml` workflow: triggered by GitHub Release, publishes `packages/stagebook` to npm
- [ ] Add `deploy-viewer.yml` workflow: builds viewer, deploys to GitHub Pages
- [ ] Update `CLAUDE.md` to reflect new structure
- [ ] Verify existing tests and build still pass after restructure

## Design decisions

- **Workspace tool**: npm workspaces (already using npm) unless there's a reason to switch to pnpm
- **Library stays the primary package**: `@deliberation-lab/stagebook` name and exports unchanged
- **Viewer imports from source** via workspace link during dev, not from published npm
[...]

PR #523, opened 2026-07-09
refactor(viewer): drop viewer/** eslint override via local type narrowing (#504)

### LINKED ISSUE #504: Tighten the viewer/** eslint override (proper typing over schema `any` fields)
Follow-up from #502 / #503.

When the preview harness was lifted from `apps/viewer` (which has **no lint script**) into `packages/stagebook/src/viewer`, it picked up the library's `recommendedTypeChecked` ESLint config. To keep #503 scoped to the move, `packages/stagebook/eslint.config.js` gained a `viewer/**` override that disables the same rules as the existing `validate/**` override, plus `no-base-to-string`:

```
"@typescript-eslint/no-unsafe-assignment": "off",
"@typescript-eslint/no-unsafe-member-access": "off",
"@typescript-eslint/no-unsafe-call": "off",
"@typescript-eslint/no-unsafe-return": "off",
"@typescript-eslint/no-unsafe-argument": "off",
"@typescript-eslint/require-await": "off",
"@typescript-eslint/no-unnecessary-type-assertion": "off",
"@typescript-eslint/no-base-to-string": "off",
```

Most of the `no-unsafe-*` hits trace back to the treatment schema typing `treatments` / `introSequences` (and similar) as `any` in the built `.d.ts`, so the fix likely spans both:

- Tighten those schema types (or add narrow local guards/annotations at the call sites) so the `no-unsafe-*` rules can be re-enabled for `viewer/**`.
- The two `no-base-to-string` sites in `StateInspector.tsx` are guarded `String(...)` calls the rule can't see through — re-typing the narrowed value would let that rule come back too.

Goal: remove the `viewer/**` override entirely (and, ideally, tackle the parallel `validate/**` override the same way). No behavior change — this is a type-safety/lint-coverage cleanup on code that was previously unlinted.

JamesPHoughton:
## What

Removes the `viewer/**` ESLint override in `packages/stagebook/eslint.config.js`. The preview harness was lifted into the linted library package (#502/#503), which surfaced `no-unsafe-*` / `no-base-to-string` hits that `apps/viewer` (no linter) never enforced.

Fixes #504.

## Approach — and why *not* the "tighten the schema types" route

The root `any` comes from `altTemplateContext` returning `z.any().superRefine(...)` — it has done so since the repo's first commit (`#61`). That `z.any()` is **load-bearing at runtime**: the schema must accept either a `{template: …}` invocation or the concrete shape and dispatch between them; the wide inferred type is a deliberate consequence. The repo already provides real types for *consuming* hydrated data via the **resolved-schema layer** (`resolved.ts`), which is the designed home for "proper TypeScript types for Stage/Element…".

I first tried the issue's primary suggestion — materialize real types for every wrapped schema. It works, but it grows the built `treatment.d.ts` from **~61 KB to ~2.5 MB** (the recursive element/stage union gets inlined at every composition point — ~288×), a cost every downstream repo's `tsc` pays. And the gain is ~zero: I checked the consumers — `annotator` imports only `stagebook/components` + `hasStableParticipantId`; `deliberation-lab` imports nothing from stagebook; nothing walks the raw treatment structure through those types. So real *raw* types would be pure cost for no consumer.

So this PR takes the issue's other stated option — **narrow locally at the viewer call sites** — keeping the foundational `any` and the ~61 KB `.d.ts`:

- **`TreatmentPicker.tsx`** — local `TreatmentSummary` / `IntroSequenceSummary` types for the two fields the picker reads.
- **`StateInspector.tsx`** — assert the non-object primitive residual for `String()` (store values are genuinely `unknown`) — `no-base-to-string`.
- **`expandTreatmentFile.ts`** — receive `fillTemplates`' `any` result as `unknown`, re-assert on return.
- **`previewResolution.test.ts`** — a local walk-type for the resolved-tree assertions.
- **`introSequencesOptional.test.tsx`** — `querySelector<T>()` instead of `as`-casts (idiomatic; also clears a `no-unnecessary-type-assertion`), one genuinely-redundant `.find()` cast dropped.

The **`validate/**` override is unchanged** — its hits are a *different* cause (traversal of the `yaml` library's untyped AST + `require-await`), tracked separately.

## Verification

- `lint` (zero warnings), `format:check`, full workspace `build` incl. `apps/viewer` `tsc -b`
- vitest 1914 (stagebook) + 78 (viewer) + 110 (vscode); Playwright CT (all browsers)
- `treatment.d.ts` stays ~61 KB (no consumer typecheck-cost regression)
- Rebased on current `main` (incl. the #520 debrief refactor)

### DIFF
--- a/packages/stagebook/eslint.config.js
+++ b/packages/stagebook/eslint.config.js
@@ -21,9 +21,14 @@ export default tseslint.config(
     },
   },
   {
-    // Validators lifted from the VS Code extension carry over their
-    // previous quality bar (apps/vscode has no linter). Tightening these
-    // is tracked for a follow-up — out of scope for the lift.
+    // Validators lifted from the VS Code extension carry over their previous
+    // quality bar (apps/vscode has no linter). Unlike the viewer lift — whose
+    // unsafe-* hits traced to the schema typing treatments/introSequences as
+    // `any` and were cleared in #504 by tightening those schema types — these
+    // hits come from a different source: traversal of the `yaml` library's
+    // loosely-typed AST (yamlPositionMap.ts) and async functions kept `async`
+    // for interface parity. Tightening them (typing the yaml AST, auditing the
+    // require-await sites) is tracked for a separate follow-up.
     files: ["**/src/validate/**/*.ts"],
     rules: {
       "@typescript-eslint/no-unsafe-assignment": "off",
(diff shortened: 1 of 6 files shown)


JamesPHoughton:
@codex review

Issue #192, opened 2026-04-22
VS Code extension: resolve asset:// URIs to local files during preview


JamesPHoughton:
## Context

#188 landed the `asset://` URI scheme for platform-provided media. In production, each deployment host (annotator, custom platform, etc.) implements its own `getAssetURL('asset://…')` — typically resolving to a presigned S3 URL or similar.

During authoring, researchers previewing treatments in VS Code can't hit that production resolver. They also may not want to — the whole reason they're using `asset://` is that the real files live in a data store separate from the repo. Currently the extension's `getAssetURL` ([apps/vscode/src/webview/index.tsx](apps/vscode/src/webview/index.tsx)) concatenates the URI onto a webview base URL, which produces a garbled path and breaks preview.

The separate viewer-placeholder issue improves the "no way to resolve it" case visually. This issue tackles the "there *is* a way — the files live on my laptop" case, which is realistic: researchers typically have the recordings/videos locally before handing them off to the platform.

## Proposal

A config-driven local mount. The researcher tells the extension "this prefix of the asset:// namespace maps to this directory on disk," and the preview uses real files.

### Option A — VS Code settings (simplest)

```jsonc
// .vscode/settings.json (per-workspace) or user settings
{
  "stagebook.assetRoots": {
    "group_recordings": "/Users/me/pilot_videos/session_01",
    "diagrams": "./study_assets/diagrams"
  }
}
```

Resolution: `asset://group_recordings/session.mp4` → look up `group_recordings` in the map → resolve `session.mp4` under `/Users/me/pilot_videos/session_01/`. Relative values resolve against the treatment file's workspace root.

Pros: familiar mechanism; per-workspace scoping falls out naturally; VS Code already surfaces settings in the Command Palette.
Cons: researcher has to discover the setting.

### Option B — sidecar file (discoverable alongside treatment)

```yaml
# stagebook.assets.yaml, adjacent to the treatment file
assetRoots:
  group_recordings: ./pilot_videos/session_01
  diagrams: ./study_assets/diagrams
```

Pros: lives with the study; shareable; researchers see it in the same directory listing as their treatment.
Cons: new file convention to document; per-user local paths get committed accidentally.

### Option C — combine

Start with (A); later allow (B) as an override if the sidecar exists. Avoid the "my paths got committed" footgun by putting the settings in `.vscode/settings.json` (git-ignored in many repos) and documenting the pattern.

[...]

PR #528, opened 2026-07-09
feat(vscode): discover asset:// prefixes in prompt bodies & field values (#524)

### LINKED ISSUE #524: VS Code preview: offer a picker for asset:// prefixes that appear only in prompt bodies / field values
Follow-up from the Codex review on #522 (#192).

The asset-mount picker banner is driven by a **host-side static scan** of the treatment YAML (`collectAssetPrefixes`), so it lists prefixes referenced by element `file`/`captionsFile` fields. Two reference paths aren't discoverable host-side and so are **not offered in the banner**:

- an `asset://` ref that appears only inside a **loaded prompt body** (`x`), and
- a **field-supplied** value (`file: ${clipUrl}` where the user enters `asset://recordings/a.mp4` in the FieldForm).

**Already handled in #522:** if such a prefix is *configured* (`stagebook.assetRoots`) or was *picked for another reference*, it still resolves — `postTreatment` advertises **every** configured/picked mount, and the `readFile` bridge resolves `asset://` prompt text from its mount. So the only gap is **interactive discovery**: an *unconfigured* prefix that appears *only* in a prompt body or a field value won't get a "Choose folder…" row.

## Proposed approach
The webview is the only place that sees every `asset://` reference at render time (via `getAssetURL`). Have it collect the prefixes that pass through unresolved and drive the banner from that set (either locally, or reported back to the extension), instead of the host-side static scan. Needs care around render-phase collection + change-detection to avoid re-render loops.

## Acceptance
- An `asset://newprefix/…` used only in a prompt body or a field value surfaces a "Choose folder…" row.
- No render loop; the banner set is stable once all referenced prefixes are mounted.

JamesPHoughton:
Fixes #524. Follow-up to #192 (#522).

## Problem
The #192 asset-mount picker banner listed only prefixes found by a **host-side static scan** of the treatment YAML. An `asset://` ref that appears only inside a loaded **prompt body** (``) or a **field-supplied value** (`file: ${clipUrl}` bound to `asset://…`) wasn't in the YAML tree, so it never got a "Choose folder…" row. (A *configured* mount still resolved it — this is only about interactive discovery.)

## Approach
The webview is the one place that sees **every** asset ref at render time, because `getAssetURL` is called for each. So:

- **`getAssetURL`** now calls `resolveAssetForRender`, which reports any ref whose prefix isn't mounted (`reportUnresolvedAsset`) and returns the resolved/passthrough URL (#192, unchanged).
- **`useAssetBanner(unmappedAssetPrefixes, assetRoots, contentVersion)`** hook:
  - collects reported prefixes into a ref **during render** (a ref write, not state), promoted to `seenUnresolved` state by a reconcile effect that **bails when unchanged** — so it **converges** (a re-render collects the same set → no further update);
  - resets the accumulation **per treatment** (`contentVersion`), during render — and resets `seenUnresolved` synchronously in the same block to avoid a one-frame stale flash;
  - returns `bannerPrefixes = mergeBannerPrefixes(...)` = union of host scan + render-collected, minus anything now mounted, sorted + deduped (stable for change-detection);
  - returns a **referentially stable** `collectUnresolved` so `getAssetURL`/`contentFns` stay stable (PreviewHost's re-fetch contract).
- The banner (`AssetMountCard`) is driven by `bannerPrefixes`; the pick flow (`pickAssetFolder`) is unchanged.

The banner still lists only prefixes seen in the *current view* plus the host scan's all-arms set; a prefix in a non-current arm's prompt body surfaces when you navigate to it.

## Testing
- **Pure helpers** unit-tested: `assetPrefixOf`, `reportUnresolvedAsset`, `resolveAssetForRender` (the getAssetURL seam), `mergeBannerPrefixes`.
- **`useAssetBanner` jsdom test** covers the load-bearing wiring: reconcile-loop **convergence** (a runaway loop would hang `act()`), discovery of a scan-missed prefix, union with the host scan, mounted-drop (both via version-reset and via the filter alone), accumulate-within-treatment, reset-on-new-treatment, and `collectUnresolved` referential stability.
- 177 vscode tests green; tsc clean (one pre-existing `semanticTokens.ts` error, unrelated); esbuild build clean.

## Review
Ran a 3-subagent review (correctness-of-React-wiring / coverage / security) with adversarial verification before opening; 4 low/nit findings folded in (stale-flash fix, getAssetURL-seam test, stability + filter-only-drop tests), 1 rejected. No new security surface — collection is render-time only; mounting/reading is the unchanged, human-gated #192 path.

### DIFF
--- a/apps/vscode/src/webview/index.tsx
+++ b/apps/vscode/src/webview/index.tsx
@@ -2,7 +2,8 @@ import React, { useState, useEffect, useMemo, useRef } from "react";
 import { createRoot } from "react-dom/client";
 import type { TreatmentFileType } from "stagebook";
 import { PreviewHost } from "stagebook/viewer";
-import { buildAssetURL } from "./resolveAsset.js";
+import { resolveAssetForRender } from "./resolveAsset.js";
+import { useAssetBanner } from "./useAssetBanner.js";
 
 // Declare the VS Code API injected by the webview
 declare function acquireVsCodeApi(): {
@@ -44,6 +45,7 @@ window.addEventListener("message", (event) => {
 function createWebviewContentFns(
   webviewBaseUri: string,
   assetRoots: Record<string, string>,
+  onUnresolvedAsset: (prefix: string) => void,
 ) {
   const cache = new Map<string, Promise<string>>();
 
@@ -66,10 +68,17 @@ function createWebviewContentFns(
     },
 
     getAssetURL(assetPath: string): string {
-      // Platform asset:// references resolve against the configured local
-      // mounts (#192); an unmapped or unsafe one passes through unchanged so
-      // the renderer shows the #191 placeholder rather than a broken URL.
-      return buildAssetURL(assetPath, webviewBaseUri, assetRoots);
+      // #524: report an asset:// ref whose prefix isn't mounted (including one
+      // that appears only in a prompt body or field value the host-side scan
+      // can't see) so the picker banner can offer it — then resolve/passthrough
+      // the URL (#192). onUnresolvedAsset is a ref write promoted to state after
+      // render (see App/useAssetBanner).
+      return resolveAssetForRender(
+        assetPath,
+        webviewBaseUri,
+        assetRoots,
+        onUnresolvedAsset,
+      );
     },
   };
 }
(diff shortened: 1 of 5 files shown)


JamesPHoughton:
@codex review

JamesPHoughton:
Closing unmerged — on reflection the premise is thinner than it looked.

- Prompt-body `asset://` refs (`` inside a `.prompt.md`) don't occur in practice.
- Field-supplied `asset://` values (`file: ${clipUrl}`) are real, but #192 (#522) **already resolves** them whenever the prefix is configured (`stagebook.assetRoots`) or previously picked — it advertises every configured mount, not just YAML-referenced ones.

So all this PR added was interactive banner *discovery* for the narrow case of a prefix that appears only in a field value **and** was never configured — at the cost of a fragile render-phase-collection React hook. Not worth carrying; #522's behavior stands. Full rationale in #524.

Issue #20, opened 2026-04-01
Accessibility initiative: audit, plan, and track a11y improvements 


JamesPHoughton:
## Purpose

Tracking issue for a mini-project to improve accessibility across the three surfaces in this repo. Not itself an actionable implementation task — the first deliverable is a plan. Subtasks get filed from that plan and linked back here.

## Why

Stagebook components run real experiments with real participants. Accessibility gaps exclude participants and contaminate the research (self-selection on ability). The current state is "whatever we happened to ship" — no explicit target standard, no audit, no testing. We should fix that intentionally rather than reactively.

## Scope (all in this monorepo)

- **`packages/stagebook`** — the component library. Participant-facing and the most important surface: form inputs, prompts, media, timeline, submit button, discussion affordances.
- **`apps/viewer`** — the standalone study previewer. Researcher-facing chrome (header, sidebar, state inspector, time scrubber, field form, landing page) plus it renders stagebook components via `Viewer`, so library gaps show up here too.
- **`apps/vscode`** — extension chrome within VS Code. Mostly inherits VS Code's own a11y story; the preview webview is the main concern since it embeds the viewer and thereby the library.

## Planning phase: deliverable

A written plan, either in this issue body (expanded below) or in `docs/decisions/` as an ADR, covering:

1. **Target standard.** WCAG 2.1 AA? 2.2 AA? Something narrower? What are we committing to and why?
2. **Audit of current state.** Run axe-core against the library components and viewer. Document violations per component/app. Manual keyboard-only walkthrough of each interactive element across the three surfaces.
3. **Gap list**, grouped by surface/component, prioritized by:
   - Severity (blocks use vs. degrades use)
   - Frequency (participant-facing vs. researcher-facing)
   - Fix cost
4. **Testing strategy.**
   - Unit / CT: per-component axe-core smoke tests in `packages/stagebook`?
   - E2E: keyboard navigation walkthrough through a full study in the viewer?
   - CI integration, flake budget
5. **Per-new-component checklist** — a short "how to ship an accessible new component" list to prevent regression.
6. **Subtask breakdown** — each gap (or cluster of related gaps) gets its own issue with acceptance criteria.

## Planning phase: non-goals

- Fixing any specific violations (those go to subtasks after the plan is agreed)
- Publishing an a11y statement / VPAT (separate concern; happens after the work)
- Perfect coverage — we're picking a standard and committing to that

## Likely out-of-scope (unless there's strong reason)

[...]

PR #546, opened 2026-07-10
fix(a11y): name the dropdown Prompt's <select> (#545)

### LINKED ISSUE #545: Prompt: dropdown renders an unnamed <select> (no accessible name)
## Bug

A `dropdown` prompt renders a `<select>` with **no accessible name** — a
participant-facing WCAG 4.1.2 / 1.3.1 violation in authored studies. Confirmed
with axe: `select-name` (critical). Surfaced by the Codex review of #541; the
dropdown is a `Prompt` variant the initial audit missed.

## Root cause

`Prompt.tsx` (dropdown branch, ~line 292) renders:

```tsx
<Select options={…} value={…} placeholder={…} onChange={…} />
```

with **no `label`**. `Select` only wires an accessible name when its `label`
prop is supplied, so the `<select>` is unnamed.

Distinct from **#538** (which adds an accessible-name *affordance* to components
that lack one): `Select` already *has* a `label` prop — the bug is that Prompt's
dropdown branch doesn't use it.

## Reproduction

axe on a dropdown `Prompt` →

```
[critical] select-name — Select element must have an accessible name
```

The other five Prompt variants pass; only `dropdown` fails.

## Fix

Give the dropdown's `<Select>` an accessible name — pass a `label` derived from
the prompt, or wire `aria-labelledby` to the rendered prompt body (mirroring how
the other Prompt variants are named). Add a CT/axe assertion for the dropdown
variant so it can't regress.

## Priority

Higher than #538 — this is critical and participant-facing (dropdown prompts
appear in real authored studies), whereas #538's TextArea case only bites hosts
using the component standalone.

## Links

- Parent: #20
- Audit: `docs/a11y-audit-2026-07.md`
- Related: #538 (form-input accessible names)

JamesPHoughton:
## What

The `dropdown` Prompt variant rendered a `<select>` with **no accessible name** — a participant-facing WCAG 4.1.2 / 1.3.1 violation (axe `select-name`, critical) in authored studies. `Select` only wired an accessible name from its `label` prop, and the dropdown branch passed none.

## Approach

Name the control by the **visible prompt body** (the question the participant reads) rather than a duplicate visible label:

- Add an `ariaLabelledBy` pass-through prop to `Select`, forwarded to the `<select>`'s `aria-labelledby`.
- Wrap the rendered prompt body in a `<div id={bodyId}>` (stable `useId()`) and point the dropdown's `aria-labelledby` at it.

The `<style>` tag `Markdown` emits inside the wrapper is `display:none`, so the accessible-name algorithm excludes it — the name resolves to the visible body text. Only the **dropdown** variant is touched; the other Prompt variants already pass axe via option-level labels.

## Tests

- New CT test asserts the combobox has an accessible name via `getByRole("combobox", { name })` (TDD red→green confirmed). **No new axe dependency added** — a consolidated axe gate is a separate later task.
- `npm test` (1980 unit tests) ✅ · `npm run build` ✅ · lint/format ✅
- Full `npm run test:ct:all`: only the pre-existing webkit `:focus-visible` focus-ring flakes fail (verified identical on `main`; CI runs with `retries: 2`), unrelated to this change.

## Scope

Minimal — dropdown variant only. Touches `Prompt.tsx`, `Select.tsx`, `Prompt.ct.tsx`; no `package.json`/lockfile changes.

Fixes #545
Refs #20

### DIFF
--- a/packages/stagebook/src/components/elements/Prompt.ct.tsx
+++ b/packages/stagebook/src/components/elements/Prompt.ct.tsx
@@ -373,6 +373,66 @@ test.describe("List Sorter", () => {
 });
 
 // ================================================================
+// Dropdown
+// ================================================================
+//
+// Inline fixture (mirrors the pattern used by the shuffle fixtures
+// below) so this file stays self-contained and doesn't touch the
+// shared fixtures module. The dropdown variant renders the prompt
+// body as a `<select>`.
+
+const dropdown = {
+  metadata: {
+    name: "projects/example/dropdownHouse.md",
+    type: "dropdown" as const,
+  },
+  body: `# Which Hogwarts house do you belong to?
+
+_Choose the house that fits you best._`,
+  responseItems: ["Gryffindor", "Hufflepuff", "Ravenclaw", "Slytherin"],
+  responsePoints: [],
+  sliderPoints: [],
+};
+
+test.describe("Dropdown", () => {
+  test("renders a select with each option", async ({ mount }) => {
+    const component = await mount(
+      <Prompt
+        {...dropdown}
+        name="testDropdown"
+        value={undefined}
+        save={() => {}}
+      />,
+    );
+    await expect(component.locator("select")).toHaveCount(1);
+    await expect(component).toContainText("Gryffindor");
+    await expect(component).toContainText("Slytherin");
+  });
+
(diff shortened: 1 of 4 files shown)


JamesPHoughton:
@codex review

JamesPHoughton:
@codex review

PR #547, opened 2026-07-10
feat(a11y): standalone accessible-name affordance for exported TextArea (#538)

### LINKED ISSUE #538: Exported form components need a standalone accessible-name affordance (TextArea)
## Purpose

Give the **exported** form components a way to receive an accessible name when
used standalone. Subtask of the accessibility initiative (#20), Step 3 of the
July 2026 component audit (`docs/a11y-audit-2026-07.md`).

## Why the export is the deciding factor

- axe flagged a standalone `TextArea` for WCAG `label` (critical): the bare
  `<textarea>` has an `id` but **no** `aria-label` / `label` / `ariaLabelledBy`
  affordance, so it has no programmatic accessible name.
- Inside Stagebook's own studies this doesn't bite — text always goes through
  `Prompt` (there is no standalone `textarea` element type), and `Prompt` names
  the textarea.
- **But `TextArea` is a public export** (`stagebook/components` →
  `components/index.ts`, `form/index.ts`) that hosts import to build their own
  participant-facing UI. A host that drops in a bare `<TextArea/>` gets an
  unnamed input — a real WCAG 1.3.1 / 4.1.2 gap, with no lever to fix it.
- This is a public-API conformance gap against **ADR Decision 1** (components
  conform to WCAG 2.2 AA self-sufficiently — not by relying on always being
  `Prompt`-wrapped, which external hosts won't do).

## The gap, precisely

`TextArea` has no label affordance (only `id`). Its siblings `RadioGroup`,
`CheckboxGroup`, and `Select` all take a `label` prop and passed the audit.

## Fix

- Add a label affordance to `TextArea` consistent with its siblings: an
  `ariaLabel` / `ariaLabelledBy` pass-through at minimum (it already takes `id`),
  ideally a `label` prop that renders a real `<label htmlFor={id}>`.
- Sweep the other exported form components to confirm each can be given an
  accessible name standalone — `TextArea` is the one the audit caught, but the
  export surface is the right lens.

## Acceptance criteria

- A standalone `<TextArea>` with a provided label/aria-label passes axe `label`.
- Each exported form input can be given an accessible name without `Prompt`
  (tested or documented).
- CT test covers the standalone-named case.

## Links

- Parent: #20
- ADR: `docs/decisions/2026-07-accessibility.md` (Decision 1)
- Audit finding: `docs/a11y-audit-2026-07.md` (Step 3)

JamesPHoughton:
## What & why

`TextArea` is a **public export** (`stagebook/components`) that hosts import to build their own participant-facing UI, outside `Prompt`. A bare `<textarea>` carried only an `id` — no `aria-label` / `<label>` / `aria-labelledby` — so used standalone it had **no programmatic accessible name**. axe flagged this for WCAG `label` (critical); it's a 1.3.1 / 4.1.2 gap and a public-API conformance gap against **ADR Decision 1** (components conform to WCAG 2.2 AA self-sufficiently, not by relying on always being `Prompt`-wrapped, which external hosts won't do).

Its siblings `RadioGroup`, `CheckboxGroup`, and `Select` all already take a `label` prop and passed the audit. `TextArea` was the odd one out.

## Change

Add three optional props to `TextArea`, mirroring the `label` prop on the siblings:

- **`label`** — renders a visible `<label htmlFor={id}>` (styled byte-for-byte like `Select`/`RadioGroup`/`CheckboxGroup`).
- **`ariaLabel` / `ariaLabelledBy`** — pass through to `aria-label` / `aria-labelledby` for the invisible-label / external-element cases.

`ariaLabel`/`ariaLabelledBy` have no default, so `undefined` renders no attribute; `label` defaults to `""` and renders no element. **The `Prompt`-wrapped path and every other existing call site are unchanged.** If more than one is supplied, the accessible-name spec resolves it deterministically (`aria-labelledby` > `aria-label` > `<label for>`).

Left untouched by design: `Prompt.tsx` / `Select` dropdown naming (that's #545).

## Sweep of the exported form inputs

Confirmed each can be given an accessible name standalone (issue's second acceptance criterion):

| Component | Affordance | Status |
|---|---|---|
| **TextArea** | `label` / `ariaLabel` / `ariaLabelledBy` | **fixed here** |
| RadioGroup | `label` → `<label htmlFor>` + `role="radiogroup"` `aria-labelledby` | already OK |
| CheckboxGroup | `label` → `<label htmlFor>` + `role="group"` `aria-labelledby` | already OK |
| Select | `label` → `<label htmlFor>` | already OK |
| Slider | range input always carries `aria-label={messages.sliderLabel}` | already OK¹ |

¹ Slider's name comes from provider messages (localizable), not a per-instance prop — it's never nameless, so no axe `label` gap. Per-instance customization would be a separate enhancement, out of scope here.

## Tests (Playwright CT — no `@axe-core/playwright`, per the issue)

`getByRole("textbox", { name })` / `toHaveAccessibleName`:

- **baseline** — a bare `<TextArea/>` has *no* accessible name (pins the exact defect + regression guard)
- `ariaLabel` names the textarea
- visible `label` prop names it **and** associates via `htmlFor`
- `ariaLabelledBy` names it from an external element

Full TextArea CT file green across chromium / firefox / webkit; `npm test` (vitest, 1980) and the build (dts) green.

## Reviews

Pre-PR correctness / test-coverage / security subagent reviews — all clean; the coverage review's one suggestion (baseline no-name test) is folded in.

Fixes #538
Refs #20

### DIFF
--- a/packages/stagebook/src/components/form/TextArea.ct.tsx
+++ b/packages/stagebook/src/components/form/TextArea.ct.tsx
@@ -2,6 +2,61 @@ import { test, expect } from "@playwright/experimental-ct-react";
 import { TextArea } from "./TextArea";
 import { MockTextArea } from "../testing/MockTextArea";
 
+// -- Standalone accessible name (#538) --
+// TextArea is a public export (stagebook/components) that hosts drop into
+// their own participant UI, outside Prompt (which otherwise names the
+// textarea). A bare <textarea> carries only an `id` — no aria-label /
+// <label> / aria-labelledby — so standalone it has no programmatic
+// accessible name, a WCAG 1.3.1 / 4.1.2 gap axe flags as `label`
+// (critical). These cover the three affordances that close it.
+
+test("standalone baseline: a bare textarea has no accessible name (the gap #538 closes)", async ({
+  mount,
+}) => {
+  // Pins the defect: with no label/aria affordance and no placeholder,
+  // the textarea's accessible name is empty. Also guards against a
+  // regression that hands it some unintended name.
+  const component = await mount(<TextArea />);
+  await expect(component.locator("textarea")).toHaveAccessibleName("");
+});
+
+test("standalone: ariaLabel gives the textarea an accessible name (#538)", async ({
+  mount,
+}) => {
+  const component = await mount(<TextArea ariaLabel="Your answer" />);
+  await expect(
+    component.getByRole("textbox", { name: "Your answer" }),
+  ).toBeVisible();
+});
+
+test("standalone: visible label prop names the textarea and associates via htmlFor (#538)", async ({
+  mount,
+}) => {
+  const component = await mount(<TextArea id="essay" label="Your answer" />);
+  // Accessible name resolves from the <label htmlFor> association.
+  await expect(
+    component.getByRole("textbox", { name: "Your answer" }),
+  ).toBeVisible();
+  // The visible <label> is rendered and points at the textarea's id.
+  const label = component.locator("label");
+  await expect(label).toHaveText("Your answer");
+  await expect(label).toHaveAttribute("for", "essay");
+});
+
+test("standalone: ariaLabelledBy names the textarea from an external element (#538)", async ({
+  mount,
+}) => {
+  const component = await mount(
+    <div>
+      <span id="q1">Describe your reasoning</span>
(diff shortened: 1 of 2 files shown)


JamesPHoughton:
@codex review

Issue #535, opened 2026-07-10
Deliberate color system for Stagebook (palette + tokens)


JamesPHoughton:
## Purpose

Choose Stagebook's color palette **intentionally**, the way the font was chosen
(Inter: widely used, robust across languages, professional, deterministic).
Today's colors are "whatever we shipped." This makes them a deliberate,
documented system. Subtask of the accessibility initiative (#20); it **subsumes**
the Step-1 contrast finding from the July 2026 component audit
(`docs/a11y-audit-2026-07.md`).

## Why color deserves the same intentionality as the font

- **It's a measurement instrument, not decoration.** Color can bias participant
  responses — an emotionally-loaded hue primes answers the same way an anchored
  slider does. A neutral, non-priming palette is a **validity** concern, not an
  aesthetic one.
- **Accessibility is a hard constraint, baked in — not a patch.** Every
  foreground/background token pairing meets WCAG 2.2 AA contrast (4.5:1 text,
  3:1 UI / large text) **by construction**. This is where the audit's 1.4.3
  finding is resolved systematically (Button / SubmitButton / TrackedLink on
  `--stagebook-primary` `#3b82f6` measure ≈3.7:1) rather than by nudging one blue.
- **Cross-cultural robustness.** Stagebook runs global studies and color carries
  culture-specific meaning (red, white, …). Never encode meaning in color alone
  (WCAG 1.4.1), and lean neutral so the instrument reads consistently across
  populations. The direct analog of "Inter works across languages."
- **Professional & restrained**, like the font — credible, doesn't editorialize
  the study.
- **Override-friendly, single canonical palette.** Keep the CSS-variable token
  architecture (`--stagebook-*`) so a study author can deliberately override the
  theme — but Stagebook ships **one** canonical, accessible default and does
  **not** auto-adapt to the participant's environment (see Non-goals).

## Deliverable

- A documented token palette — base/surface, text (primary/secondary/muted),
  interactive (primary + hover/active), semantic (success/warning/danger),
  border/focus — **one canonical set**.
- Every documented pairing contrast-validated at WCAG 2.2 AA.
- `color-scheme: light` pinned and native controls audited, so a participant's OS
  dark mode can't re-tint native form controls / scrollbars out from under the
  study.
- A short rationale doc: the "why these colors" companion to the Inter decision.

## Acceptance criteria

- Resolves the Step-1 contrast finding: Button, SubmitButton, and TrackedLink
  pass axe `color-contrast` at AA.
[...]

Issue #494, opened 2026-07-01
Automate VS Code extension packaging/publishing (committed .vsix is stale)


JamesPHoughton:
## Problem

`apps/vscode/stagebook-vscode-0.1.0.vsix` is committed to the repo but hand-maintained: it was first added in #482 (i18n) and has **never been rebuilt since**, so the tracked artifact is missing at least:

- #483 / #491 — editor + viewer prompt locale-consistency
- #442 — "Stagebook: Validate Workspace" command

Anyone installing the extension from the committed `.vsix` gets a build several features behind `main`. There's currently:

- **No publishing pipeline** — `npm-publish.yml` publishes the *library* to npm on release; `deploy-viewer.yml` deploys the viewer; `ci.yml` only tests/builds. Nothing runs `vsce publish` or rebuilds the `.vsix`.
- **No documented install path** — the extension isn't on the VS Marketplace / OpenVSX, there's no `apps/vscode/README.md`, and no docs tell users how to install it. The committed `.vsix` is the de-facto (undocumented) channel.

## Options to decide between

1. **Publish to a registry (recommended to evaluate first).** Add a `vsce publish` (VS Marketplace) and/or `ovsx publish` (OpenVSX) step, triggered on a version tag or GitHub release — mirroring `npm-publish.yml`. Users install by name; no binary in git. Requires a marketplace publisher account + `VSCE_PAT`/`OVSX_PAT` secrets, and an `apps/vscode/README.md` + icon for the listing.
2. **Keep committing the `.vsix`, but rebuild it in CI.** A workflow (or a check) that regenerates the `.vsix` from `main` so it can't drift — e.g. attach it as a GitHub Release asset, or fail CI if the committed `.vsix` is out of date. Avoids a marketplace account but keeps a binary in the release flow.
3. **Untrack the `.vsix`.** If distribution should happen via (1), remove it from git + `.gitignore` it so it stops drifting in the meantime.

## Immediate vs. long-term

Regardless of the long-term choice, `main` currently ships a stale `.vsix`. Decide whether to do an interim "rebuild + commit" (or remove) now, or wait for the automation.

## Acceptance

- A documented, reproducible way for study-repo users to install a current build of the extension.
- The distributed artifact can't silently fall behind `main` (published on release, rebuilt in CI, or untracked).
- Version-bump story defined (the extension is stuck at `0.1.0`; decide whether it tracks the library release or versions independently).

## Context

[...]

Issue #298, opened 2026-05-07
Bake position into reference grammar (breaking)


JamesPHoughton:
## Motivation

Position is currently a sibling field on every condition leaf:

```yaml
- reference: prompt.recall.value
  position: 0
  comparator: equals
  value: red
```

But position is *part of the lookup*, not a modifier — it's "whose copy of this value." Putting it inside the reference makes references self-describing strings, simpler to log, easier to use as inputs to aggregates (where every input is a fully-qualified ref), and consistent with how mongo, JSONPath, and several declarative DSLs handle scoped lookups.

## Proposed grammar

The first segment of a dotted reference is the position selector; the second is the source enum:

```yaml
# explicit position
- reference: 0.prompt.recall.value          # player at position 0
- reference: self.prompt.notes.value        # current player's own value
- reference: shared.prompt.consensus.value  # group state
- reference: all.prompt.recall.value        # multi-position read (where supported)
- reference: any.aggregate.partisanScore.value
```

Structured form (#240) gets a `position:` field, mirroring the prefix:

```yaml
- reference:
    position: 0
    source: prompt
    name: recall
    path: [value]
```

## Breaking change

**This issue intentionally requires the position prefix on every reference**, rather than defaulting to `self`/`player` when omitted. Reasons:

- The migration window is small now (handful of studies). It grows over time.
- Implicit defaults are footguns — author forgets to specify and gets data from the wrong scope. Recent stagebook history (e.g. percentAgreement removal, the storage-key reuse tightening in #281) has favored explicit-over-implicit when the safety win is bigger than the verbosity cost.
- A bare `prompt.X.value` becomes ambiguous now anyway — it could be parsed as `<source>.<name>.<path>` (today) or `<position>.<source>.<name>` (with the new prefix). Forcing the prefix removes the ambiguity from the parser.

## What changes

- **Parser** in [packages/stagebook/src/schemas/reference.ts](packages/stagebook/src/schemas/reference.ts): first segment is a position selector (numeric integer, `self`, `shared`, `all`, `any`); second is the source enum. Reject references that don't lead with a position.
- **Structured form** gets a required `position:` field.
- **Sibling `position:` field on condition leaves** is removed (the prefix carries the info; keeping both creates two ways to say the same thing).
- **Validation** in [validateReferences.ts](packages/stagebook/src/schemas/validateReferences.ts) walks the new prefix.
- **Resolved schemas** (resolved.ts) inherit the new shape.

## Examples — before / after

[...]

PR #559, opened 2026-07-11
fix(preview): timeline reload + AA-palette drift in viewer chrome & VS Code webview

### LINKED ISSUE #558: Timeline selections don't survive a stage reload (missing self. prefix after #298)
## Bug

The `timeline` element reads back previously-saved selections so a participant who reloads a stage sees their existing marks. That read is built at runtime in [`Element.tsx`](packages/stagebook/src/components/Element.tsx) as `resolve(\`timeline.${name}\`)` — **without the `self.` position prefix** that #298 made mandatory.

Because the reference is malformed, `resolve()` catches it, logs `Invalid reference: "timeline.<name>" — … missing a position prefix` to the console, and returns `[]`. So `initialSelections` is `undefined` and the saved marks silently vanish on reload. The write path is fine (saves under `timeline_<name>`); only the read-back is broken.

The sibling Prompt path one screen up already does it correctly (`self.prompt.<name>`). This was a #298-migration miss.

## Impact

- Console error on every timeline render (noise).
- **Timeline is the one element whose saved state doesn't survive a reload** — a reproducibility bug for studies using timeline selections.

## Why it wasn't caught

The reference is synthesized at runtime, so `stagebook validate` never sees it, and no test round-tripped a saved timeline selection through the real post-#298 resolver.

## Fix

Add the `self.` prefix: `resolve(\`self.timeline.${name}\`)`, plus a jsdom test that saves a selection and asserts it's read back into `initialSelections` (closes the coverage gap). Found while investigating a stale VS Code preview.

JamesPHoughton:
## Summary

Started as "why does my VS Code preview show old-blue buttons but a correct rose-700 playhead?" and uncovered two real bugs plus a third stale chrome surface.

### The root cause of the "timeline right, buttons old" split
The VS Code webview **hand-maintains a copy** of the stagebook `:root` design tokens (`extension.ts` `getWebviewContent`) because the bundled `styles.css` is loaded as text, not auto-injected — so that copy is the webview's *only* source of `:root` tokens. It still pinned `--stagebook-primary` to the **retired blue-500**, overriding the whole preview to old-blue, while `--stagebook-playhead` (absent from the copy) correctly fell back to the component default rose-700. That's exactly the split observed. This was invisible because the source components and the standalone browser preview were already correct.

## Changes

**Participant-facing correctness (`Fixes #558`)**
- `Element.tsx` read the timeline's saved selections via the pre-#298 bare `timeline.<name>` reference (no position prefix), so `resolve()` rejected it, logged `Invalid reference`, and returned `[]` — **timeline marks silently vanished on stage reload**. Added the mandatory `self.` prefix. The write path was fine; only the read-back was broken. The sibling Prompt path already did this correctly.

**Preview color drift (#535 follow-through, `refs #494`)**
- Synced the webview's injected `:root` tokens to the current palette (blue-600/700/800; `text-faint` → `decoration`).
- Swept the retired blue-500 out of every chrome surface → `var(--stagebook-primary, #2563eb)`: the reusable harness (Next / FieldForm / TimeScrubber / SkeletonPlaceholder + webview card button) **and** the standalone viewer app (App / OverviewPage / LandingPage).
- Bumped the extension `0.1.0` → `0.1.1` so VS Code recognizes the new build (same-version installs don't reload).

**Guards (the gap that let this drift — `styles.test.ts` only scanned the stylesheet)**
- `webviewTokenSync.test.ts` asserts every color token the webview hand-copies **equals what `styles.css` resolves it to** — catches the *next* drift, not just the one retired hex (verified: fails on a mutated token).
- `chromeColorDrift` (viewer lib + viewer app) and `webviewColorDrift` (extension) fail if the retired blue-500 reappears; all recursive.

Also preserves the #535 color audit as `docs/color-system-audit-2026-07.html`.

## Testing
- vitest **2030** (stagebook) + **168** (vscode, incl. value-sync) + **79** (viewer) ✓
- Playwright CT (chromium): green (1 known timing flake, passes in isolation) ✓
- Rebuilt + reinstalled the `0.1.1` vsix; verified the bundle contains **zero** retired blue-500 and the injected `--stagebook-primary` is now blue-600.

## Pre-PR review (3 subagents)
- **Security:** clean — the `self.` prefix + `position==="self" → "player"` mapping fully constrain the reference to the current participant (no cross-participant leakage from a crafted `name`); no CSP/injection surface.
- **Correctness:** no defects — round-trip verified end-to-end; every injected token matches `styles.css`.
- **Coverage:** the test is a real regression gate (verified red without the fix); its findings (value-sync test, viewer-app sweep, recursive guards) are folded in above.

### DIFF
--- a/apps/viewer/src/App.tsx
+++ b/apps/viewer/src/App.tsx
@@ -375,7 +375,8 @@ const buttonStyle: React.CSSProperties = {
   padding: "0.5rem 1rem",
   borderRadius: "0.375rem",
   border: "none",
-  backgroundColor: "#3b82f6",
+  // Accent token (fallback = blue-600) so app chrome tracks the palette.
+  backgroundColor: "var(--stagebook-primary, #2563eb)",
   color: "white",
   cursor: "pointer",
   fontSize: "0.875rem",
--- a/apps/viewer/src/chromeColorDrift.test.ts
+++ b/apps/viewer/src/chromeColorDrift.test.ts
@@ -4,13 +4,14 @@ import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
 
 // Drift guard (#535). `#3b82f6` is the retired blue-500 that used to be
-// `--stagebook-primary`. Viewer chrome hardcoded it in inline styles and
-// silently kept it through the #535 bump to blue-600 (#2563eb), so the preview
-// rendered a mix of old and new blue: participant components tracked the token,
-// the chrome didn't. styles.test.ts already guards the stylesheet, but the
-// chrome lives in inline `.tsx` styles it never scanned — this closes that gap.
-// Chrome must reference `var(--stagebook-primary, #2563eb)` (or a deliberate
-// non-accent color), never the retired literal.
+// `--stagebook-primary`. The standalone viewer app is a third chrome surface
+// (alongside the reusable harness in packages/stagebook and the VS Code
+// webview) that hardcoded the retired blue-500 in inline styles and kept it
+// through the #535 bump to blue-600. App chrome must reference
+// `var(--stagebook-primary, #2563eb)` (the app imports stagebook/styles, so the
+// token is defined; the fallback is blue-600 too), never the retired literal.
+const srcDir = dirname(fileURLToPath(import.meta.url));
+
 function walk(dir: string): string[] {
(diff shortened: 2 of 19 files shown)


JamesPHoughton:
@codex review

Issue #560, opened 2026-07-11
VS Code preview should render components with the library's real styles (stop hand-copying tokens & resets)


JamesPHoughton:
## Problem

The VS Code preview webview (`apps/vscode/src/extension.ts` → `getWebviewContent`) **hand-maintains a parallel copy of the library's visual styling** in an inline `<style>` block, instead of using the library's own CSS. Because the bundled `styles.css` is loaded as text (`--loader:.css=text`) and never injected, that hand-copy is the webview's *only* source of `:root` tokens and several resets.

This is DX-breaking: the preview is how we visually inspect components during development, so any divergence between the preview and the real library rendering makes the preview **actively misleading**, not just cosmetically off.

It already bit us — see #559 (root cause of the "timeline right, buttons old" bug): the hand-copied `--stagebook-primary` was pinned to the retired blue-500 while the real palette had moved to blue-600, so every `var(--stagebook-primary)` in the preview rendered old-blue. #559 corrected the values and added a value-sync guard (`webviewTokenSync.test.ts`), but that's a workaround for a copy that shouldn't exist.

## What's duplicated in `getWebviewContent`

| Block | Also owned by |
|---|---|
| `:root` `--stagebook-*` token block | `packages/stagebook/src/styles.css` `:root` |
| `input[type=checkbox/radio]` + `:checked` SVG data URIs | component inline styles (RadioGroup/CheckboxGroup/Markdown, per #213) |
| `code` / `pre` styling | Markdown component |
| `table` / `th` / `td` styling | Markdown component (#214) |
| form-field font/border resets (implicit) | `styles.css` form resets |

Everything except genuinely webview-specific chrome (`--viewer-sidebar-width`, `body { margin: 0 }`, the sidebar layout) is a reimplementation of styling the library already ships.

## Proposed approach

1. **Inject the library's real `styles.css`** into the webview. The bundle already has it as text (`--loader:.css=text`); the webview entry (`src/webview/index.tsx`) imports that text and injects a single `<style>` on startup (CSP already allows `style-src 'unsafe-inline'`).
2. **Delete the hand-copied token block and duplicated resets** from `getWebviewContent`, keeping only webview-specific chrome.
3. Once the copy is gone, **delete `webviewTokenSync.test.ts`** (the sync guard has nothing to sync) — replace it with a test asserting the webview injects `styles.css`.

Net effect: the preview renders every component with the exact styles the library ships, so drift is structurally impossible — the property we actually want.

[...]

PR #563, opened 2026-07-11
feat(vscode): derive the extension version from the stagebook version

### LINKED ISSUE #562: Tie the VS Code extension version to the bundled stagebook version (major/minor, build-time-derived)
## Problem

The VS Code extension's version is independent of the stagebook library it bundles: the extension currently reads `0.1.2` while it bundles stagebook `0.24.x`. So two useful questions have no checkable answer:

- **"Is my installed extension current?"** — nothing ties the extension number to anything meaningful.
- **"Which stagebook am I previewing against?"** — invisible; the extension version doesn't reflect the bundled library at all.

This was a real source of friction (see #559/#560/#561: the extension version being frozen at `0.1.0` across rebuilds is what made "have I actually updated the preview?" unanswerable and let a stale build linger).

## Decision

Mirror the bundled stagebook's **major/minor** in the extension version; let the **patch float** for extension-only fixes. Since stagebook is pre-1.0, that means:

```
extension version = 0.<stagebook-minor>.<extension-patch>
```

- **major/minor** = the bundled stagebook's feature set (a real compatibility signal — patches within a minor are backward-compatible, so a small patch mismatch is safe)
- **patch** = the extension's own counter, free for webview/validation fixes between stagebook releases

Exact-match (`extension == stagebook`) reads cleaner but degrades to this the moment there's an extension-only fix (no free number to bump), so major/minor + floating patch is the stable choice.

## Must be build-time-derived, not hand-maintained

A hand-copied version number is the **same failure mode** as the hand-copied design tokens #560 just removed. So:

- A build step reads `packages/stagebook/package.json` `version` and stamps the extension's **major/minor** to match; the extension bundles that same stagebook, so they're aligned by construction — drift impossible.
- **Patch rule:** if stagebook's minor changed since the last extension build, reset the extension to `0.<new-minor>.0`; if the minor already matches, leave the patch alone (author bumps it manually for extension-only fixes). Simple and monotonic.
- Home: the `cut-release` flow and/or the extension publish pipeline (#494 — the "no publish pipeline, hand-maintained vsix" gap this dovetails with).

## Explicitly NOT doing: an in-preview version stamp

Considered and declined. The VS Code Extensions panel already shows the installed version, and once the version tracks stagebook, that answers "am I current." A persistent version badge in a measurement-instrument preview is exactly the kind of chrome the design ethos avoids. If exact-bundled-patch debugging is ever needed, a zero-footprint `console.log("stagebook <version>")` on webview startup is the clutter-free way — not chrome. Hold even that until there's a concrete need.

## Acceptance criteria

- [ ] A build/release step derives the extension major/minor from `packages/stagebook/package.json`; a stagebook minor bump that isn't reflected fails the check (or auto-syncs).
- [ ] Patch semantics documented (reset on minor change; manual bump for extension-only fixes).
- [ ] No hand-maintained version number; no in-preview stamp.
- [ ] Wired so cutting a stagebook release keeps the extension in step.

## Related
- #494 — VS Code extension publish pipeline (natural home for the automation)
- #559 / #560 / #561 — the drift saga this addresses the "which build am I running?" tail of

JamesPHoughton:
## Summary

Closes the "which build am I running?" tail of the #559/#560/#561 preview-drift saga. The VS Code extension previews against the stagebook it bundles, but its version was independent (`0.1.x`) — so "which stagebook is this?" and "is my extension current?" had no answer, and a frozen `0.1.0` is what let a stale build linger unnoticed.

`Fixes #562`

## What changed

**Derive the version, never type it** (a hand-maintained number is the same drift trap the design-token copy was in #560):

- **`scripts/sync-vscode-version.mjs`** — sets `apps/vscode/package.json`'s version to the `packages/stagebook` version. Idempotent (no-op when equal); fail-closed (aborts if it can't find the field).
- **`prepackage` hook** — runs it before every `vsce package`, so both a local `npm run package` and the `release-vscode.yml` CI job emit a vsix whose version mirrors stagebook.
- **`versionSync.test.ts`** — guards the committed values from drifting a minor.
- **`cut-release` skill** — the release bump now runs the sync + `npm install --package-lock-only`, kept as a version-only, auto-mergeable diff; "don't hand-edit the extension version" guard added.
- **`release-vscode.yml`** — corrected the now-false "version does not advance" comment.

Because the extension only ships alongside a stagebook release (`release-vscode.yml` fires on `release: published`), exact-match is the natural result; major/minor is the compatibility-meaningful part.

## Verified end-to-end
Forced the extension version to `0.1.2`, ran `npm run package` → `prepackage` re-synced to `0.24.0` and vsce emitted `stagebook-vscode-0.24.0.vsix`. So `vsce` reads the version **after** `prepackage`. Also confirmed: idempotent no-op leaves the tree clean; the drift-guard test fails on a minor mismatch; the release flow is a closed loop (`npm ci` + `versionSync` both fail-closed if the sync is skipped).

## Effect
Cutting stagebook `0.25.0` now automatically ships the extension as `0.25.0` — the Extensions panel answers "which stagebook / am I current?" at a glance. No chrome, no manual step. (Deliberately did **not** add an in-preview version stamp — the panel + version-tying covers it; see #562 for that decision.)

## Review
Pre-PR correctness review: sound, no defects (script regex/fail-path, cwd-independent path resolution, CI/release-flow interaction all verified empirically). Security surface is negligible — a build script that rewrites one version field, no runtime/participant/auth surface.

### DIFF
--- a/.claude/skills/cut-release/SKILL.md
+++ b/.claude/skills/cut-release/SKILL.md
@@ -33,6 +33,10 @@ Note: prior versions are not strictly sequential — `0.10.2` was skipped. Don't
 git checkout main && git pull origin main
 git checkout -b release/X.Y.Z
 # edit packages/stagebook/package.json: bump "version" field only
+# Mirror the version onto the VS Code extension so the installed extension
+# reports which stagebook it bundles (#562). This is derived, not typed:
+node scripts/sync-vscode-version.mjs   # sets apps/vscode/package.json to X.Y.Z
+npm install --package-lock-only        # reflect both bumps in the lockfile
 git commit -am "chore(release): bump stagebook to X.Y.Z
 
 <release notes — same shape as the PR body in step 3>
@@ -41,7 +45,7 @@ Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>"
 git push -u origin release/X.Y.Z
 ```
 
-The commit subject must follow the format `chore(release): bump stagebook to X.Y.Z` — the GitHub Action keys off this.
+The commit subject must follow the format `chore(release): bump stagebook to X.Y.Z` — the GitHub Action keys off this. The diff is version bumps only (stagebook + the derived extension version + the lockfile) — still no code to review.
 
 ## 3. Open the release PR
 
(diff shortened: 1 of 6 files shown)


JamesPHoughton:
@codex review

PR #581, opened 2026-07-23
docs: shared-notepad defaultText is placeholder-only

JamesPHoughton:
Fixes #579

## What

Records the semantics decision from talkbench/runner#311: for `shared: true` open-response prompts, the `defaultText` passed to the `renderSharedNotepad` slot is **placeholder-only** — ephemeral hint text that disappears once anyone types, never seeded into the shared document, never part of the saved/exported value. This matches non-shared open response, where the prompt file's `> ` lines render as the textarea's HTML `placeholder`.

- **docs/researcher/elements.md** — shared-notepad example now shows a `> ` placeholder line (in a proper third `---` section — the previous example was a two-section file the parser rejects) and states the semantics. The corrected example validates clean via `stagebook validate - --type=prompt`.
- **docs/engineer/platform-requirements.md** — replaces the "Support default text initialization" requirement (which implied seeding) with the placeholder-only contract; services line now mentions the TalkBench runner's Yjs/CodeMirror coedit alongside Etherpad.
- **docs/engineer/architecture.md**, **docs/engineer/api-reference.md** — correct the stale `{ padName }`-only slot signature to the real `{ padName, defaultText?, rows? }` (already shipped on main; docs catching up).
- **docs/engineer/integration-guide.md** — updated example destructuring all three config fields + host contract paragraph.
- **StagebookProvider.tsx** — JSDoc-only: states the contract on the slot for IntelliSense. No behavioral change.

## Review

Three pre-PR subagent reviews (correctness / test coverage / security) ran against this diff:

- Correctness caught the invalid researcher example (missing `---` response-section delimiter) — fixed and validated before opening.
- Test coverage: no tests needed for a docs/JSDoc diff; the researcher example's shape is already covered by `promptFile.test.ts` fixtures. One gap filed as follow-up: #580 (nothing pins the Prompt → `renderSharedNotepad` config pass-through the docs now promise).
- Security: no findings.

Local: lint + all vitest workspaces green. Playwright CT not run — markdown + comment-only diff, no runtime surface.

Related: talkbench/runner#311 (placeholder implementation), talkbench/runner#421 (caption → "Shared" chip)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #583, opened 2026-07-24
test: pin the Prompt → renderSharedNotepad config pass-through

JamesPHoughton:
Fixes #580

## What

Adds `Prompt.test.tsx` (jsdom) pinning the host contract documented in #581's docs: a `shared: true` openResponse prompt calls the `renderSharedNotepad` slot with exactly `{ padName, defaultText, rows }`, where `defaultText` is the prompt file's `> ` lines newline-joined. Three cases:

- shared → slot called with the exact config, slot content renders, no textarea
- multi-line placeholder → `defaultText` is `"\n"`-joined (a single-item fixture can't distinguish `join("\n")` from `responses[0]`)
- non-shared → slot never called, textarea renders

Exact-object matching is intentional contract-pinning: adding or renaming a key in the slot config should fail this test.

## Why jsdom, not Prompt.ct.tsx

Playwright CT proxies function props across the mount boundary as async message channels — a render prop can't hand JSX back to the browser-side component, so the slot never renders under CT (verified empirically; it's also why the existing CT shared-mode test can only assert the textarea is absent). The jsdom form follows the `ImageElement.test.tsx` harness conventions and gives a real `vi.fn()` spy.

## Verification

- Guard verified by mutation: temporarily dropping `defaultText` from the call site makes the test fail; restoring goes green.
- Pre-PR subagent reviews (correctness / coverage+security): correctness confirmed the assertion is deterministic (the settled last call always carries the threaded response items; openResponse never shuffles) and flagged two inert excess props, dropped; coverage suggested the newline-join case, added; security clean.
- Local: lint + all vitest workspaces green (2,382 tests); full `test:ct:all` run had 2 chromium Timeline failures that pass in isolation (unrelated interaction-test flakes; this diff is a jsdom test only).

### DIFF
--- a/packages/stagebook/src/components/elements/Prompt.test.tsx
+++ b/packages/stagebook/src/components/elements/Prompt.test.tsx
@@ -1,0 +1,101 @@
+// @vitest-environment jsdom
+import { describe, test, expect, vi, afterEach, beforeAll } from "vitest";
+import React from "react";
+import { act } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { Prompt } from "./Prompt.js";
+import { openResponse } from "./fixtures/prompts.js";
+
+beforeAll(() => {
+  (
+    globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }
+  ).IS_REACT_ACT_ENVIRONMENT = true;
+});
+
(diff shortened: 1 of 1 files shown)


JamesPHoughton:
@codex review

